PR #60, opened 2019-07-12
167002452 testing

jesusgonzalezrivera:


santostiago:
@jesusgonzalezrivera I'm having 661 examples and 193 failures. Any idea why?

app/models/fmu_operator.rb
```diff
@@ -98,7 +96,7 @@ def self.calculate_current
   # Updates the list of documents for this FMU
   def update_documents_list
     current_operator = self.fmu.operator.id rescue nil
-    return unless self.operator.fa_id.present?
+    return if self.fmu&.operator&.fa_id.blank?
```
santostiago: @jesusgonzalezrivera why?

app/models/observation.rb
```diff
@@ -71,8 +73,8 @@ class Observation < ApplicationRecord
 
   validates :country_id,       presence: true
   validates :publication_date, presence: true
-  validates_presence_of :validation_status
-  validates_presence_of :observation_type
+  validates :validation_status, presence: true
```
santostiago: @jesusgonzalezrivera why?

app/models/operator.rb
```diff
@@ -210,17 +212,20 @@ def calculate_scores
...
+        .order("score_absolute OFFSET #{2 * third_operators}").update_all(score: 3)
     end
   end
 
 
   def rebuild_documents
-    return if fa_id.blank?
-    country = RequiredOperatorDocument.where(country_id: country_id).any? ? country_id : nil
+    # TODO: Refactor, use the same logic that appears on create_documents
+    return if fa_id.blank? || country_id.blank?
```
santostiago: @DimaZab we cannot change the code to this.
The logic is the following:
There are required documents that have no country id. This required documents should be required for countries that have no required documents themselves (for example, if we were to calculate it for Ukraine, since there are no required documents for Ukraine, then this operator would have the required documents that have no country id associated)

app/models/operator.rb
```diff
@@ -258,7 +261,7 @@ def create_operator_id
   def create_documents
     return if fa_id.blank? || country_id.blank?
 
-    country = RequiredOperatorDocument.where(country_id: country_id).any? ? country_id : nil
+    country = RequiredOperatorDocument.where(country_id: country_id).exists? && country_id
```
santostiago: @DimaZab same here (we should change the previous line to be the same as it was in `rebuild_documents`

app/models/operator.rb
DimaZab: ok, will fix it

app/models/operator.rb
DimaZab: ok

app/models/required_operator_document.rb
```diff
@@ -30,16 +30,11 @@ class RequiredOperatorDocument < ApplicationRecord
   has_many :operator_document_countries
 
   validates :valid_period, numericality: { greater_than: 0 }
-  after_destroy :invalidate_operator_documents
```
santostiago: @jesusgonzalezrivera why did you remove this?

app/models/fmu_operator.rb
jesusgonzalezrivera: My mistake, I just wanted to use `blank?` instead of `unless` for making it more understandable. it is necessary to remove the `.fmu&.`:
```ruby
return if self.operator&.fa_id.blank?
```

app/models/required_operator_document.rb
jesusgonzalezrivera: When I removed it, only god and me... It is because the relation with `operator_documents` is defined with `dependent: :destroy`. So if we remove a `RequiredOperatorDocument`, we are also removing the `operator_documents` and it doesn't make sense to update it status.

app/models/user_permission.rb
```diff
@@ -100,7 +100,7 @@ def role_permissions
             gov_file: { rud: { gov_document: { required_gov_document: { country_id: user.country_id  }}}, create: {}}
         }
       else
-        { user: { id: user.id }, observations: { read: {} } }
+        { user: { current: { id: user.id }, read: { id: user.id } }, observations: { read: {} } }
```
santostiago: @DimaZab what does the `current` key do here? This will be read by the cancan ability and it won't know what to do with that key, right?

config/initializers/extensions_ignore_tables.rb
```diff
@@ -0,0 +1,3 @@
+# https://stackoverflow.com/questions/40882420/rails-postgis-error-cannot-drop-table-spatial-ref-sys
```
santostiago: @DimaZab maybe we should change the `db:schema:load` no to crash.

db/schema.rb
```diff
@@ -634,13 +634,6 @@
     t.index ["severity_id"], name: "index_severity_translations_on_severity_id", using: :btree
   end
 
-  create_table "spatial_ref_sys", primary_key: "srid", id: :integer, force: :cascade do |t|
```
santostiago: @jesusgonzalezrivera 
1 - Why did you remove this?
2 - There's one migration missing in here, the one that creates tutorials.
3 - After removing all the migrations, did you create a task to update the migrations table? How did you manage it? (maybe this is in a file I haven't yet seen).

spec/acceptance/shared_examples/jsonapi_resources/create.rb
```diff
@@ -0,0 +1,46 @@
...
+      it "Returns error object when cannot be created by admin" do
+        post(route,
+             params: jsonapi_params(@collection, nil, options[:invalid_params]),
+             headers: success_headers)
+
+        expect(parsed_body).to eq(jsonapi_errors(*options[:error_attributes]))
+        expect(status).to eq(422)
+      end
+
+      it "Returns success object when was seccessfully created by admin" do
```
santostiago: @DimaZab typo in `successfully`

spec/acceptance/shared_examples/jsonapi_resources/delete.rb
```diff
@@ -0,0 +1,29 @@
...
+
+RSpec.shared_examples 'jsonapi-resources__delete' do |options|
+  context "Delete" do
+    let(:resource) { create(@singular.to_sym) }
+    let(:route) { "/#{@route_key}/#{resource.id}" }
+    let(:success_headers) { options[:success_role] ? authorize_headers(create(options[:success_role]).id) : webuser_headers }
+    let(:failure_headers) { options[:failure_role] ? authorize_headers(create(options[:failure_role]).id) : webuser_headers }
+
+    describe 'For admin user' do
+      it "Returns success object when was seccessfully deleted by admin" do
```
santostiago: @DimaZab typo

spec/acceptance/shared_examples/jsonapi_resources/edit.rb
```diff
@@ -0,0 +1,43 @@
...
+      it "Returns error object when cannot be updated by admin" do
+        patch(route,
+              params: jsonapi_params(@collection, resource.id, options[:invalid_params]),
+              headers: success_headers)
+
+        expect(parsed_body).to eq(jsonapi_errors(*options[:error_attributes]))
+        expect(status).to eq(422)
+      end
+
+      it "Returns success object when was seccessfully updated by admin" do
```
santostiago: @DimaZab same here

app/models/user_permission.rb
DimaZab: It's for our [v1/users_controller#current](https://github.com/Vizzuality/fti_api/blob/master/app/controllers/v1/users_controller.rb#L9) action. I can change it to check directly in action like `can :read, resource`

spec/acceptance/shared_examples/jsonapi_resources/create.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+RSpec.shared_examples 'jsonapi-resources__create' do |options|
+
+  context "Create" do
+    let(:route) { "/#{@route_key}" }
+    let(:success_headers) { options[:success_role] ? authorize_headers(create(options[:success_role]).id) : webuser_headers }
```
santostiago: @DimaZab maybe we could change the `success_role` and `failure_role` to arrays to be able to test it with different roles other than `user` and `admin`. Or we can do it at the controller spec level. Or we could have a test just for permissions. What do you think?

DimaZab: Yes, we can. Good idea to make them an array.

spec/acceptance/v1/operators_spec.rb
```diff
@@ -1,194 +1,58 @@
...
-    end
+      let(:operator) { create(:operator) }
 
-    context 'Delete operators' do
       describe 'For admin user' do
-        before(:each) do
-          token    = JWT.encode({ user: admin.id }, ENV['AUTH_SECRET'], 'HS256')
-          @headers = @headers.merge("Authorization" => "Bearer #{token}")
-        end
+        it 'Upload logo and returns success object when the operator was seccessfully updated by admin' do
```
santostiago: @DimaZab typo in "successfully"

DimaZab: Sorry, all these typos come from master, I've copied it from there. But forgot to check them.

spec/acceptance/v1/user_registarion_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'rails_helper'
```
santostiago: @DimaZab there's a typo in the file name.

spec/acceptance/v1/user_registarion_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        password: 'password',
+        password_confirmation: 'password',
+        permissions_request: 'government',
+        nickname: 'sebanew',
+        name: 'Test user new'
+      }
+    end
+
+    describe 'Registration' do
+      it 'Returns error object when the user cannot be registrated' do
```
santostiago: @DimaZab typo in the work `registered`.

spec/factories/governments.rb
```diff
@@ -9,14 +9,10 @@
...
+FactoryBot.define do
   factory :government do
-    government_entity 'A Government'
-    details           'Indicator one'
-
-    after(:create) do |government|
-      government.update(country: FactoryGirl.create(:country, name: "Country #{Faker::Lorem.sentence}",
-                                                              iso: "C#{Faker::Lorem.sentence}"))
-    end
+    country
```
santostiago: @DimaZab shouldn't we keep consistency and use: `country ||= FactoryBot.create(:country)` ?

spec/factories/observation_documents.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :observation_document do
+    user
```
santostiago: @DimaZab same here.

spec/factories/observations.rb
```diff
@@ -27,56 +27,54 @@
...
-      annex = FactoryGirl.create(:annex_operator)
-
-      observation.update(severity: FactoryGirl.create(:severity, severable: annex),
-                         annex_operator: annex,
-                         user: FactoryGirl.create(:admin),
-                         observer: FactoryGirl.create(:observer, name: "Observer #{Faker::Lorem.sentence}"),
-                         operator: FactoryGirl.create(:operator, name: "Operator #{Faker::Lorem.sentence}"),
-                         species: [FactoryGirl.create(:species)])
-    end
+    severity
```
santostiago: @DimaZab same here

spec/factories/operator_documents.rb
```diff
@@ -0,0 +1,38 @@
+FactoryBot.define do
+  factory :operator_document do
+    user
```
santostiago: @DimaZab same

spec/factories/operators.rb
```diff
@@ -25,14 +25,11 @@
 #  approved                           :boolean          default(TRUE), not null
 #
 
-FactoryGirl.define do
+FactoryBot.define do
   factory :operator do
-    name "Operator #{Faker::Lorem.sentence}"
+    country
```
santostiago: @DimaZab same

spec/models/fmu_operator_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+            expect(another_fmu_operator.valid?).to eql false
+            expect(another_fmu_operator.errors[:current]).to eql(
+              ['There can only be one active operator at a time']
+            )
+          end
+        end
+      end
+    end
+
+    describe '#non_clliding_dates' do
```
santostiago: @jesusgonzalezrivera typo

spec/models/fmu_operator_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    it { is_expected.to belong_to(:operator).required }
+  end
+
+  describe 'Validations' do
+    # We have a before_validation which sets start_date, so we cant test this
+    #it { is_expected.to validate_presence_of(:start_date) }
+
+    describe '#start_date_is_earlier' do
+      context 'when end_date is present' do
+        context 'when start_date is greather or equal to the end_date' do
```
santostiago: @jesusgonzalezrivera typo

spec/models/fmu_operator_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+        another_fmu = create(:fmu, geojson: { properties: {} })
+        @activate_fmu_operator = create(
+          :fmu_operator,
+          current: false,
+          fmu: another_fmu,
+          start_date: Date.today,
+          end_date: Date.tomorrow)
+      end
+
+      it 'deactive old fmu_operators and activate current ones with new properties' do
```
santostiago: @jesusgonzalezrivera typo in `deactivate`

spec/models/operator_spec.rb
```diff
@@ -28,46 +28,373 @@
...
+        Operator.all.map { |operator| operator.calculate_observations_scores }
+
+        5.times do |t|
+          country = create(:country)
+          operator = create(:operator, country: country, is_active: true, fa_id: "fa_#{t}")
+          operator.update_attribute(:score_absolute, t / 10.to_f)
+        end
+      end
+
+      it 'update score of the operators ordering by score and spliting then into 3 groups' do
```
santostiago: @jesusgonzalezrivera typo in `splitting`

spec/models/user_permission_spec.rb
```diff
@@ -13,71 +13,114 @@
 require 'rails_helper'
 
 RSpec.describe UserPermission, type: :model do
-  before :each do
-    @user     = create(:user)
-    @user_ngo = create(:user, permissions_request: 'ngo')
+  before :all do
```
santostiago: @jesusgonzalezrivera and @DimaZab since the `abilities_spec.rb` was removed, shouldn't we have some tests here that actually test the abilities of the users, instead of just testing the hash of permissions?

spec/factories/governments.rb
DimaZab: I haven't seen any reason for it.

DimaZab: Make sense

config/initializers/extensions_ignore_tables.rb
DimaZab: Great, but I think it would be valid only for the development environment.

DimaZab: It will take more time than I think. Because this solution only for dumps. For now, I think we can skip this step.

spec/models/user_permission_spec.rb
DimaZab: Makes sense

app/models/operator.rb
```diff
@@ -256,7 +259,7 @@ def create_operator_id
   end
 
   def create_documents
-    return if fa_id.blank? || country_id.blank?
+    return if fa_id.blank?
```
santostiago: @DimaZab this should stay as it was before.
We don't want to create documents for operators that have no country_id.
What we want is to create documents for operators who have a country_id but whose country has no required documents.

app/models/required_operator_document.rb
santostiago: @jesusgonzalezrivera but because we're using `paranoia` they're not removed from the database, and they should still be invalidated. Correct?

db/migrate/20190617164035_create_tutorials.rb
```diff
@@ -1,11 +0,0 @@
-class CreateTutorials < ActiveRecord::Migration[5.0]
```
santostiago: @jesusgonzalezrivera this file cannot be deleted.

santostiago: @jesusgonzalezrivera ok, I'll solve it some other way.

db/migrate/20190617164842_create_uploaded_documents.rb
```diff
@@ -1,12 +0,0 @@
-class CreateUploadedDocuments < ActiveRecord::Migration[5.0]
-  def change
```
santostiago: @jesusgonzalezrivera this should not have been removed either. I'll fix it.

app/models/required_operator_document.rb
santostiago: I've just re-added it.

PR #123, opened 2020-02-19
Feature/appsignal

santostiago:
The goal of this PR is to add and set up appsignal for this project, so that we can monitor if something goes wrong in staging/production without the need to access the server.

In order to do so, it was necessary to upgrade ruby and to regenerate the gemfile.lock, and some tests are not passing and should be fixed before merging. It's also important to check the backoffice is working properly, as we have no tests for it yet.

@DimaZab could you take a look at this later, please?


app/models/operator_document.rb
```diff
@@ -141,7 +144,7 @@ def delete_previous_pending_document
                                                fmu_id: self.fmu_id,
                                                required_operator_document_id: self.required_operator_document_id,
                                                status: OperatorDocument.statuses[:doc_pending])
-    pending_documents.each {|x| x.destroy}
+    pending_documents.each { |x| x.destroy }
```
DimaZab: I notice that is a light refactoring, but you can use `.each(&:destroy)`

app/models/operator_document.rb
```diff
@@ -28,8 +28,8 @@
 class OperatorDocument < ApplicationRecord
   acts_as_paranoid
 
-  belongs_to :operator, required: true, touch: true
-  belongs_to :required_operator_document, ->() { with_archived }, required: true
+  belongs_to :operator, optional: false, touch: true
+  belongs_to :required_operator_document, -> { with_archived }, required: true
```
DimaZab: `optional: false`?

app/models/user_permission.rb
```diff
@@ -26,82 +26,82 @@ def change_permissions
...
-        { user: { manage: { id: user.id } } ,
-          operator_document: { manage: { operator_id: user.operator_id } },
-          operator_document_annex: { ud: { operator_document: { operator_id: user.operator_id } }, create: {} },
-          observation: { read: {} },
-          fmu: { ru: {} },
-          operator: { ru: { id: user.operator_id } },
-          sawmill: { create: {}, ud: { operator_id: user.operator_id } } }
-      when 'ngo'
-        { user: { manage: { id: user.id } },
+  def role_permissions
```
DimaZab: I think better to move all permissions hashes to constants. It will reduce method length and provide the ability to use this hashes in the future with `UserPermission::ADMIN_PERMISSIONS` for example.

PR #125, opened 2020-02-24
Changes the way the user is associated to an observation using the importer

santostiago:
The `user_id` that the observations created through the importer should not be in the CSV.
Instead, it should be grabbed from the the user authentication key used to perform the request.


app/controllers/v1/imports_controller.rb
```diff
@@ -24,6 +24,10 @@ def importer
       @importer ||= FileDataImport::BaseImporter.build(import_params[:importer_type], import_params[:file])
     end
 
+    def user_id
```
DimaZab: Better to name it like `user_id _params` or `importer_extra_attributes`, because of `user_id` method name associated with a number.

PR #129, opened 2020-02-26
Fixed sawmill geojson query.

DimaZab:


santostiago:
@DimaZab I switched the function `json_create_object` with `jsonb_create_object`. Isn't that approach better?

PR #131, opened 2020-02-28
Feature/vector tiles

santostiago:
It creates an endpoint to generate vector tiles.

It still needs tests and maybe performance optimization (for example by using a `prepare` SQL statement, but it needs to be in develop now and later we'll improve on it)


santostiago:
(still fixing the specs)

PR #153, opened 2020-03-26
Observation filters endpoint should take params

santostiago:
The observations filters endpoint would return all the filters with no parameters.
This was changed to accept the same parameters as the observations endpoint.


santostiago:
Tests are missing.

PR #226, opened 2021-05-19
Fix/couple fixes

tsubik:
Fixing and re-enabling couple specs. I thought simplecov is broken when I saw 7% coverage, but maybe it is true? :fearful: 

I'm not sure if I dind't break anything, when fixing specs, look at my comments.


config/routes.rb
```diff
@@ -4,7 +4,7 @@
   devise_for :users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self) rescue ActiveAdmin::DatabaseHitDuringLoad
 
-  root to: "home#index"
+  root to: redirect('/admin')
```
tsubik, line 7: `home#index` looked more like on purpose, but I'm not sure. I wonder if that won't break anything, don't know how this is setup along with frontend portals.

app/models/fmu.rb
```diff
@@ -148,15 +148,19 @@ def update_geojson
...
     temp_geojson['properties']['fmu_type_label'] = Fmu::FOREST_TYPES[self.forest_type.to_sym][:geojson_label] rescue ''
+
+    self.geojson = temp_geojson
   end
 
   def update_properties
-    if self.operator.present?
-      self.properties['company_na'] = self.operator.name
-      self.properties['operator_id'] = self.operator.id
+    self.properties = {} if properties.blank?
```
tsubik, line 156: it was erroring in tests, as properties does not have default `{}` value. That would error for new FMUs.

app/models/fmu.rb
```diff
@@ -148,15 +148,19 @@ def update_geojson
     temp_geojson['properties']['certification_ls'] = self.certification_ls
     temp_geojson['properties']['observations'] = self.active_observations.reload.uniq.count
     temp_geojson['properties']['fmu_type_label'] = Fmu::FOREST_TYPES[self.forest_type.to_sym][:geojson_label] rescue ''
+
+    self.geojson = temp_geojson
```
tsubik, line 152: This seems to be removed by accident. Is running update on FMUs on production needed to correct this mistake? To backfill recently changed geojsons?

app/models/fmu.rb
```diff
@@ -56,15 +56,15 @@ class Fmu < ApplicationRecord
   accepts_nested_attributes_for :operators
   accepts_nested_attributes_for :fmu_operator, reject_if: proc { |attributes| attributes['operator_id'].blank? }
 
-  before_validation :update_geojson
+  before_validation :update_geojson, :update_properties
```
tsubik, line 59: I put it into the same place as other properties update. For consistency

app/models/score_operator_document.rb
```diff
@@ -66,8 +66,8 @@ def replace(operator)
   # We also remove the one whose required_operator_documents have been deleted
   # @param [RequiredDocumentsQuery] query_builder the query method to use
   def calculate_scores(query_builder)
-    self.all = ScoreOperatorPresenter.new(self).all
-    #self.all = query_builder.new.call(operator.operator_documents.non_signature).count.to_f / RequiredDocumentsQuery.new.call(operator.operator_documents.non_signature).count.to_f
+    # self.all = ScoreOperatorPresenter.new(self).all
```
develaper, line 69: that was the good one

config/routes.rb
develaper: I agree that looked like on purpose but never tried. If you have the chance to test it locally with the front end and it keeps working for me is ok. I f not, don't worry to much. It seems that it has been working properly like that for a long while so we can let it like that.

app/models/fmu.rb
develaper: We have kind of an epic task related with fmus not updating properly.
You will se that a _fmu_ has _properties_ and also has _geojson_ and I am affraid that here we are doing nothing here because all those properties should be in _self.properties_ instead of _self.geojson_

tsubik: I don't understand why there are two different attributes. But I believe this geojson should stay, even if new properties are now also saved.

config/routes.rb
tsubik: Going to leave it as it is, as I don't have local setup for otp portal, would require more time to setup, but that would have to be tested on staging, maybe later.

app/models/fmu.rb
develaper: Just adding some info here to avoid miscommunication. 
Currently I am working on those 2 tasks
https://www.pivotaltracker.com/story/show/178175997
https://www.pivotaltracker.com/story/show/178176207
That are linked to the whole fmus assignment and update processes and we may end up overlapping changes here.

tsubik: Ahh, ok. So what should I do with those changes? good to merge?

tsubik: So you can continue with your changes after mergin that in.

develaper (approved):

develaper:
hi @tsubik ! Would you be able to take a look to the failing test please?
rspec ./spec/acceptance/operators_spec.rb:32 # Operators GET /operators Listing operators
For the rest I am super ok.
Thanks

tsubik:
Hmm, that looks like flaky test @develaper . Would be hard now to find the reason, probably something was not cleaned up correctly by DatabaseCleaner, I don't see that factories could create those 2 extra operators there. I just hit rebuild, I wouldn't worry about that now.

PR #228, opened 2021-05-21
cleaning up operator document uploading

tsubik:
Couple changes:

- updates DocumentFileUploader to use operator_document model for generating the name
- removed old OperatorDocumentUploader
- create history was fired on every save, even if there were no changes. That didn't make sense, beside after_save is fired twice when saving operator_document, not sure if not why becasue of saving file_document also, dunno. But what I already now this version of Rails is buggy, was trying after_commit with previous_changes in lambda and it does not work.

What's left maybe for a new task
- remove attachment column in operator_document table, I think that's not needed anymore, ideas? The same for `file_name` in DocumentFile, it is reduntant now as name is generated, even if provided the file_name is in attachment column.
- run a script to rename files on production, to get rid of `attachment.pdf`


tsubik:
I'm not sure what was the reason to migrate to separate table DocumentFile, now seems like it is more complicated.

develaper (approved):

PR #230, opened 2021-05-27
Fix/regenerate operator document files

tsubik:
In this PR:
- main objective is the one off rake task to fix bad names of attachments which were created over the last couple weeks. Script by default run in dry run mode, meaning nothing will be updated, just will return a bunch of useful info about the state of attachments. To run script to do the thing append env variable `FOR_REAL=true`
- removing old `attachment` column from `OperatorDocument` as attachments were migrated to separate `DocumentFile` model.
- removing redundant `file_name` column from `DocumentFile`, as the attachment column is the same.

### DIFF
--- a/app/models/document_file.rb
+++ b/app/models/document_file.rb
@@ -6,7 +6,6 @@
 #
 #  id         :integer          not null, primary key
 #  attachment :string
-#  file_name  :string
 #  created_at :datetime         not null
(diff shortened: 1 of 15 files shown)


tsubik:
This should not be run before merging the fix of attachment names to prod.

develaper (approved):

PR #233, opened 2021-05-31
Chore/GitHub actions

tsubik:


tsubik:
Trying to fix FFI problem without ruby upgrade, but still we need to think about it.

develaper (approved):

PR #234, opened 2021-06-01
Fixes fmu properties update

develaper:
**UPDATE!! 09/07/2021:**
Forget about those 2 branches that I was talking about yesterday!
I have discarded removing duplicated "properties". It was not going to work for the front end. Then properties is updated, then geojson[" properties"] is merged with properties updated.
I have changed the query in update_geometry to work only over the current fmu. I don't see any reason to go over all the fmus neither.

**UPDATE END**

**UPDATE!! 08/07/2021:**
I decide to split the code that was here into two different branches.
First one, this one:

- will update the Fmu.properties column to match the relationship reflected in the fmu_operator table so properties will have company_na and operator_id.

**spoiler**: we don't need to sync the info at geojson.properties neither at geometry because we are not going to use it anymore
Removes properties from `def update_geometry`

why? First of all because it was messing around mismatching properties. 
As far as I understand, we are adding to the geometries information that is already present in the geojson column.
This info is duplicated in properties and some of the information also is reflected at its own column or directly accessible:
`'id', 'fmu_name', 'iso3_fmu', 'company_na', 'operator_id', 'certification_fsc', 'certification_pefc', 'certification_olb', 'certification_pafc', 'certification_fsc_cw', 'certification_tlv', 'certification_ls', 'observations', 'fmu_type_label'`

- SO, the second PR's purpose is to get rid of that annoying duplication of information deleting duplicated data from geometries, geojson and properties but retrieving the same info with the same structure to avoid any unexpected error somewhere somehow.

**UPDATE END**

Here I am trying to fixes the lack of sync between the information stored in the properties of some Fmus(forest management units).
_fmu.rb_
`has_one :fmu_operator, ->{ where(current: true) }`
`has_one :operator, through: :fmu_operator`

and its properties should look like this:
`properties['company_na'] = operator&.name`
`properties['operator_id'] = operator&.id`

I have added this task to spot "wrong" fmus and save them:
lib/tasks/fmus.rake
Saving the object triggers some clalbacks:
This one always updates geojson['properties'] so we will always call update_geometry:
`before_validation :update_geojson`
I have modified this one to ensure that the info in properties match the info in fmu_operator:
`before_save :update_properties`
And this one is **_a little motherfucker_**:
`after_save :update_geometry, if: :geojson_changed?`
Why?
I have discovered that after running rails fmu:current FOR_REAL=true the wrong fmus were fixed but 2 fmus that were OK get broken.
After a session of debugging I am 99% sure that nothing unexpected happens until, somehow, running _update_geometry_ for the broken fmus brokes the properties for "some" fmus that are OK.
```
def update_geometry
    query =
      <<~SQL
        WITH g as (
        SELECT *, x.properties as prop, ST_GeomFromGeoJSON(x.geometry) as the_geom
        FROM fmus CROSS JOIN LATERAL
        jsonb_to_record(geojson) AS x("type" TEXT, geometry jsonb, properties jsonb )
        )
        update fmus
        set geometry = g.the_geom , properties = g.prop
        from g
        where fmus.id = g.id;
      SQL
    ActiveRecord::Base.connection.execute query
  end
```
If this callback is removed or simply commenting the content of the method makes all the fixing work as expected and no other fmu is harmed in the process.
SO, my biggest doubt here is why do we use this query instead of a simple update_attributes approach? And how can this query affect to a different fmu?


app/models/fmu.rb
```diff
@@ -156,14 +157,10 @@ def update_geojson
...
-      properties['company_na'] = operator.name
-      properties['operator_id'] = operator.id
-    else
-      properties['company_na'] = nil
-      properties['operator_id'] = nil
-    end
+    properties['company_na'] = operator&.name
+    properties['operator_id'] = operator&.id
+    
+    self.properties
```
tsubik, line 163: Do you need to return properties here?

tsubik (commented):
Hmm, I wonder why All geometeries are updated when geojson changed and not just the changed one.

tsubik:
Maybe executing that query on ActiveRecord::Base in after_save running in not the same transaction. Maybe moving `update_geometry` to `after_commit` will helped, but then `geojson_changed?` won't be available, there is `previous_changes` hash for this.

tsubik:
Also when using `after_commit` and `previous_changes` the hash could not provide correct changes in `if` statement, when used with lambda like this `-> { previous_changes['foo'].present? }` just saing that this Rails version is buggy, and it worked for me when using `proc` instead, or checking the change in the callback message.

app/models/fmu.rb
```diff
@@ -132,38 +130,29 @@ def esri_shapefiles_zip=(esri_shapefiles_zip)
...
-    temp_geojson['properties']['certification_olb'] = self.certification_olb
-    temp_geojson['properties']['certification_pafc'] = self.certification_pafc
-    temp_geojson['properties']['certification_fsc_cw'] = self.certification_fsc_cw
-    temp_geojson['properties']['certification_tlv'] = self.certification_tlv
-    temp_geojson['properties']['certification_ls'] = self.certification_ls
-    temp_geojson['properties']['observations'] = self.active_observations.reload.uniq.count
-    temp_geojson['properties']['fmu_type_label'] = Fmu::FOREST_TYPES[self.forest_type.to_sym][:geojson_label] rescue ''
-
-    self.geojson = temp_geojson
+  def properties
```
tsubik, line 133: I would rename that to `geojson_properties`. I hope that will be performant enough, but yeah this is aggressive caching, so maybe it will be ok

app/models/fmu.rb
```diff
@@ -132,38 +130,29 @@ def esri_shapefiles_zip=(esri_shapefiles_zip)
...
+    temp_properties['company_na'] = self.operator&.name
+    temp_properties['operator_id'] = self.operator&.id
+    temp_properties['certification_fsc'] = self.certification_fsc
+    temp_properties['certification_pefc'] = self.certification_pefc
+    temp_properties['certification_olb'] = self.certification_olb
+    temp_properties['certification_pafc'] = self.certification_pafc
+    temp_properties['certification_fsc_cw'] = self.certification_fsc_cw
+    temp_properties['certification_tlv'] = self.certification_tlv
+    temp_properties['certification_ls'] = self.certification_ls
+    temp_properties['observations'] = self.active_observations.reload.uniq.count
```
tsubik, line 148: I'm not sure if that will work correctly with our aggressive caching, `active_observation` count can change, but the cached FMUs will still return old value. There could be millions problems like this in this codebase actually, this is not related with just this task.

tsubik: Maybe we should periodically clean up cache.

tsubik (changes requested):
I think I saw usage of `properties` also in `calculate_current` fmu_operator model. That needs to be removed as well.

app/models/fmu.rb
```diff
@@ -208,16 +197,15 @@ def update_geometry
...
-        SELECT *, x.properties as prop, ST_GeomFromGeoJSON(x.geometry) as the_geom
+        SELECT *, ST_GeomFromGeoJSON(x.geometry) as the_geom
         FROM fmus CROSS JOIN LATERAL
         jsonb_to_record(geojson) AS x("type" TEXT, geometry jsonb, properties jsonb )
         )
         update fmus
-        set geometry = g.the_geom , properties = g.prop
+        set geometry = g.the_geom
         from g
         where fmus.id = g.id;
```
agnessa, line 207: As far as I can tell this query affects all fmus, not sure why that would be needed after an update of an individual records, but it answers the question why other records are affected. I imagine the reason it is called this way is to make use of the lateral join, which allows to easily call ST_GeomFromGeoJSON and update in a single query. But why for all records - no idea.

app/models/fmu.rb
develaper: nope. Thanks

app/models/fmu.rb
develaper: After checking a few times the code base I couldn't find a place where the properties of the geometry can be updated. Meaning that the api gets a request or from the admin panel. Those properties that are only reflected in the geometry.properties, like centroid, were set when the geometry was imported into the DB. Since then `def update_geometry` is called only in the callback . At this callback we only update properties that are actually already reflected in fmu.properties. That's why I don't see the point at all in updating the geometry properties. What do you think?

tsubik: I also don't know why this query is affecting all fmus. Should that be only for the updated one?

app/models/fmu.rb
```diff
@@ -156,14 +157,8 @@ def update_geojson
...
     self.properties = {} if properties.blank?
-
-    if operator.present?
-      properties['company_na'] = operator.name
-      properties['operator_id'] = operator.id
-    else
-      properties['company_na'] = nil
-      properties['operator_id'] = nil
-    end
+    properties['company_na'] = operator&.name
```
tsubik, line 160: I wonder if we cannot just merge update_properties with update_geojson, and save ALL properties, just saving in two places.

app/models/fmu.rb
tsubik: Is there any other place to update properties? I think there is not, only `update_properties` callback, but it only updates 2 of them, so I think you can remove it, but you need to update all properties like geojson in that callback.

tsubik (changes requested):

app/models/fmu.rb
develaper: I have discarded removing duplicated "properties". It was not going to work for the front end. Then properties is updated, then geojson[" properties"] is merged with properties updated.
I have changed the query in update_geometry to work only over the current fmu. I don't see any reason to go over all the fmus neither.

agnessa (approved):

tsubik:
One test is failing otherwise looks good :+1:

PR #235, opened 2021-06-01
fixing total countries and calculation of ranking operator documents

tsubik:
This also requires to fix some data in the database.


tsubik:
My plan is to remove RankingOperatorDocuments table altogether as we don't need history data for the ranking. Instead I will just compute ranking on demand as this is very simple and performant SQL query like this
```
SELECT 
operators.id as operator_id, 
country_id,
CASE 
  WHEN "all" = 0 THEN
    COUNT(*) OVER (PARTITION BY country_id)
  ELSE 
    RANK() OVER (PARTITION BY country_id ORDER BY "all" DESC) 
END as position,
COUNT(*) OVER (PARTITION BY country_id) as total
FROM score_operator_documents
  INNER JOIN operators on operators.id = score_operator_documents.operator_id
  AND operators.fa_id <> 'NULL'
  AND operators.is_active = true
  AND score_operator_documents.current = true
```

tsubik:
Needs fixed score operator documents table data first.

app/resources/v1/operator_resource.rb
```diff
@@ -105,8 +105,7 @@ def country_doc_rank
     end
 
     def country_operators
-      country_operators = @model.country&.operators
-      country_operators&.count
+      @model.ranking_operator_document&.total
```
develaper, line 108: I am bit confused about this method.
Where does the .total comes from and what does it returns? I see it in RankingOperatorDocument >> attr_accessor but I can't find where is set or defined.

tsubik: Total is total number of operators in the country here `COUNT(*) OVER (PARTITION BY o.country_id) as total`. And is passed to this object in constructor RankingOperatorDocument.new(ranking). Me being lazy, I just used ActiveModel::Model to have many active record features.

develaper (approved):

PR #236, opened 2021-06-07
Fix/score operator document unique constraint

tsubik:
The data on prod should be fixed first otherwise it won't run this migration.


tsubik:
Well, other way could be to add `current_score_operator_document_id` into `operators` table, but to be consistent doing such a refactor I would do the same for observations, also that would be more work to do that kind of refactor.

develaper (approved):

PR #239, opened 2021-06-14
Fix/fixing scores recalculate script

tsubik:


tsubik:
We can add scheduler task to run `sync:scores_last_week` every day. Just to fix any errors if there are. Maybe a materialized view that would calculate all scores based on histories table could be better, but the query would be complicated and every refresh would take couple seconds and in the future more.

tsubik:
Add contract signature filter, fixed sort in operator document histories index page, and also display all ScoreOperatorDocument entities (chart raw data) in operator show page, but for now only in development mode.

tsubik:
Ranking tests are flaky, but that is for another task.

PR #245, opened 2021-06-24
Fix/changes to scores

tsubik:
Changing how scores are calculated and documents returned to the public

- ALL private documents are returned as `not_provided` if no signed contract
- score calculation changes because of that


tsubik:
I'm adding more tests for ScoreOperatorPresenter

develaper (approved):

PR #259, opened 2021-08-03
Feature/Dashboards

tsubik:


tsubik:
It is confirmed to be OK by client.

tsubik:
False alarm, only one dashboard of three was approved. Anyway, I do not expect big changes in the remaining two.

develaper (approved):

PR #274, opened 2021-10-12
Feature/observation responsible admin

develaper:
https://www.pivotaltracker.com/story/show/174314841

### DIFF
--- a/app/models/observer.rb
+++ b/app/models/observer.rb
@@ -104,6 +104,7 @@ class Observer < ApplicationRecord
   # Sets the default responsible admin for an observer
   # 
   def set_responsible_admin
+    return if self.responsible_admin.present?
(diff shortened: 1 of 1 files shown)


app/models/observer.rb
```diff
@@ -98,6 +101,12 @@ def cache_key
     super + '-' + Globalize.locale.to_s
   end
 
+  # Sets the default responsible admin for an observer
+  # 
+  def set_responsible_admin
+    self.responsible_admin = User.where(email: ENV['RESPONSIBLE_EMAIL'].downcase).first
```
tsubik, line 107: I think it should only set default if no one select any other user. Could be misleading to create monitor and overlook that responsible admin is not the one that was selected.

tsubik (changes requested):
I think only a small issue with UX of creating new monitor.

PR #278, opened 2021-10-17
Fix/document expire notifications

develaper:
- Added a task to Send notification for expiring documents that will be call using a cronjob. _**Note to myself:Remember to create it after deploying to production.**_
- Changed notify_operator_expiring_document to notify when "expired today" and "expiring soon" but only to those that have operator.email.


app/services/expiration_notifier_service.rb
```diff
@@ -5,7 +5,7 @@ class ExpirationNotifierService
 
   # @param [Date] date The date for which to notify the users
   def initialize(date = Date.today)
-    @notification_dates = NOTIFICATION_PERIODS.map { |x| date - x }
+    @notification_dates = NOTIFICATION_PERIODS.map { |x| date + x }.push(Date.yesterday)
```
tsubik, line 8: Maybe it would be better to add [-1.day] to NOTIFICATION_PERIODS?

tsubik: Great spot with that `+` sign lol

app/services/expiration_notifier_service.rb
```diff
@@ -16,9 +16,14 @@ def call
...
-        documents = OperatorDocument.active.where(expire_date: date, operator_id: operator.id)
-        SendExpirationEmailJob.perform_later(operator, documents)
+      OperatorDocument.where(expire_date: date).select(:operator_id).group(:operator_id).each do |operator_document|
+        documents = OperatorDocument.where(expire_date: date, operator_id: operator_document.operator_id)
+        operator = Operator.find(operator_document.operator_id)
+        # We need an email adress to send the email
+        # Almost all the operator have email == nil
+        # we could use the document.user.email but
+        # also a huge number of document.user_id are nil
+        SendExpirationEmailJob.perform_now(operator, documents) unless operator.email == nil
```
tsubik, line 26: why not just `operator.email.nil?`

tsubik: or better if `operator.email.present?`

app/jobs/schedule_expiration_emails_job.rb
```diff
@@ -4,7 +4,6 @@ class ScheduleExpirationEmailsJob < ApplicationJob
   queue_as :scheduled_tasks
```
tsubik, line 4: I think this job is not used anywhere, maybe we can remove it?

develaper: i have to confess that I was affraid

PR #287, opened 2021-11-02
Fix/expiring notification multilang

develaper:


app/services/mail_service.rb
```diff
@@ -186,25 +186,54 @@ def notify_responsible(observation)
...
+    if expire_date == Date.today + 1.month
       # expiring documents
-      time_to_expire = distance_of_time_in_words(expire_date, Date.today)
-      subject = t('backend.mail_service.expiring_documents.title', count: num_documents) +
-        time_to_expire
-      text = [t('backend.mail_service.expiring_documents.text', company_name: operator.name, count: num_documents)]
-      text << time_to_expire
-      documents.each { |document|  text << "<br><a href='#{document_admin_url(document)}'>#{document&.required_operator_document&.name}</a>" }
+      I18n.locale = :en
+      subject = t('backend.mail_service.expiring_documents.title.month')
```
tsubik, line 192: I'm wondering if maybe just normal action mail templates could be used instead of this.

tsubik (approved):

tsubik:
Please, fix the specs first.

app/services/mail_service.rb
develaper: I couldn't find a straight forward way to add the templates and make it work. Then searched for some examples and realized that there is no email template used in the whole project. I guess that there should be a reason but did not had time to investigate further.

PR #291, opened 2022-04-19
user update with password, when changing password

tsubik:
When changing email or password, the current password should be mandatory.

OTP portal does not have profile update yet, but observation tool does, so to not break profile update there I will temporarily allow password change without current password unless it is provided as a parameter. Later after changing observation tool that should be removed.

### DIFF
--- a/app/controllers/v1/users_controller.rb
+++ b/app/controllers/v1/users_controller.rb
@@ -8,16 +8,17 @@ module V1
 
     def update
       # this is temporary thing allow transition for observation tool
-      if context[:app] == 'observations-tool' && !update_params[:current_password]
+      if use_json_api_resources?
         super
       else
-        user = User.find(context[:current_user])
+        user = User.find(params[:id])
         if user.send(update_action, update_params)
           render json: JSONAPI::ResourceSerializer.new(
             UserResource
           ).serialize_to_hash(UserResource.new(user, context))
         else
-          render json: ErrorSerializer.serialize(user.errors, 422), status: :unprocessable_entity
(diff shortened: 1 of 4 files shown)


tsubik:
Tested with observation tool locally connected.

app/controllers/v1/users_controller.rb
```diff
@@ -6,11 +6,64 @@ class UsersController < ApiController
 
     load_and_authorize_resource class: 'User'
 
+    def update
+      # this is temporary thing allow transition for observation tool
+      if use_json_api_resources?
```
santostiago, line 11: I'm a fan of using guard clauses, like
```
return super if use_json_api_resources?
```

But all good :D

app/controllers/v1/users_controller.rb
```diff
@@ -6,11 +6,64 @@ class UsersController < ApiController
...
+    end
+
+    def update_action
+      return "update_with_password" if validate_current_password?
+
+      "update"
+    end
+
+    def validate_current_password?
+      update_params[:password] || update_params[:email]
```
santostiago, line 48: We're going to allow them to modify the email as well?

santostiago (approved):
Looks good!!

app/controllers/v1/users_controller.rb
tsubik: We will remove it soon.

app/controllers/v1/users_controller.rb
tsubik: Disallow to modify email without providing current_password

PR #298, opened 2022-05-23
Feat dynamic about page

santostiago:
Changes the BE to support dynamic about pages.
[Task](https://www.pivotaltracker.com/story/show/181285333)

## Pull request checklist

- [ ] Check the commit's or even all commits' message styles matches our requested structure.

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/app/admin/about_page_entries.rb
+++ b/app/admin/about_page_entries.rb
@@ -13,9 +13,9 @@ ActiveAdmin.register AboutPageEntry do
 
   filter :position, as: :select
   filter :translations_title_contains, as: :select, label: 'Title',
-          collection: AboutPageEntry.with_translations(I18n.locale).pluck(:title)
+                                       collection: AboutPageEntry.with_translations(I18n.locale).pluck(:title)
   filter :translations_body_contains, as: :select, label: 'Body',
-         collection: AboutPageEntry.with_translations(I18n.locale).pluck(:body)
+                                      collection: AboutPageEntry.with_translations(I18n.locale).pluck(:body)
 
   controller do
     def scoped_collection
--- a/app/admin/about_page_entries.rb
+++ b/app/admin/about_page_entries.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register AboutPageEntry do
+
+  extend BackRedirectable
+  back_redirect
+
+  menu false
+
+  config.order_clause
+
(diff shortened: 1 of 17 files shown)


app/admin/about_page_entries.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register AboutPageEntry do
```
tsubik, line 3: Not sure if it wouldn't be better to have `Page` model with `sections`. Something similar we have here https://github.com/Vizzuality/laws_and_pathways/blob/develop/app/models/page.rb#L20

tsubik: But we can always refactor this if there will be need, dunno.

app/admin/about_page_entries.rb
```diff
@@ -0,0 +1,80 @@
...
+              }
+    end
+    f.actions
+  end
+
+  show do
+    attributes_table do
+      row :position
+      row :title
+      row :body
```
tsubik, line 73: I would do `html_safe` to not show encoded html tags.

app/admin/about_page_entries.rb
```diff
@@ -0,0 +1,80 @@
...
+    column :title
+    column :body
+    column :created_at
+    column :updated_at
+  end
+
+  index do
+    column :position
+    column :title
+    column :body
```
tsubik, line 37: Not sure if needed on index page, gonna take some space.

tsubik: But I see it is in other places. I would do `html_safe` everywhere.

tsubik (approved):
Small changes to use `html_safe` to show the real content in show/index pages.

app/admin/about_page_entries.rb
santostiago: @tsubik to have more than one page with dynamic content? 
Yeah, there might be the need in the future, but since we don't have any for these 2 contracts it might be premature optimization.
If we end up needing another one, we can refactor this :)

app/admin/about_page_entries.rb
santostiago: @tsubik there are not that many entries. Showing the body should be fine. But we can remove it if we later see it's occupying too much space.

PR #299, opened 2022-05-28
Make sawmill name mandatory

tsubik:
and adding missing specs for API.

### DIFF
--- a/app/models/sawmill.rb
+++ b/app/models/sawmill.rb
@@ -17,16 +17,16 @@
 
 class Sawmill < ApplicationRecord
   belongs_to :operator
+
+  validates_presence_of :name
   validates_numericality_of :lat, greater_than_or_equal_to: -90, less_than_or_equal_to: 90
   validates_numericality_of :lng, greater_than_or_equal_to: -180, less_than_or_equal_to: 180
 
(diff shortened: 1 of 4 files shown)


app/models/sawmill.rb
```diff
@@ -17,17 +17,17 @@
...
 class Sawmill < ApplicationRecord
   belongs_to :operator
+
+  validates_presence_of :name
   validates_numericality_of :lat, greater_than_or_equal_to: -90, less_than_or_equal_to: 90
   validates_numericality_of :lng, greater_than_or_equal_to: -180, less_than_or_equal_to: 180
 
   scope :active, -> { Sawmill.where(is_active: true) }
   scope :inactive, -> { Sawmill.where(is_active: false) }
+  scope :filter_by_operators, ->(operator_ids) { where( operator_id: operator_ids.split(',')) }
```
santostiago, line 27:
```suggestion
  scope :filter_by_operators, ->(operator_ids) { where( operator_id: operator_ids) }
```

You don't need to convert to a string :)

santostiago (approved):
All good! 👍

app/models/sawmill.rb
tsubik: I was just removing extra space from there, not touching that.

PR #302, opened 2022-06-02
Setup Sentry

tsubik:


config/initializers/sentry.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sentry.init do |config|
```
santostiago, line 3: Shouldn't we configure this only for production?

santostiago (approved):

tsubik:
No, it should be everywhere. Just don't set env variables in development
and it doesn't report. Should be on staging for sure I will set env
variables on servers

PR #307, opened 2022-06-15
Add operator_id parameter to fmu tiles

tsubik:
Parameter `operator_id` would be great to just fetch FMU tiles that belongs to operator (for FMUs map). Otherwise we download all tiles, wasting server resources and there is a bug that I'm trying to reproduce that is showing shape from FMU that does not even belong to viewed operator.

### DIFF
--- a/app/admin/fmu.rb
+++ b/app/admin/fmu.rb
@@ -70,6 +70,28 @@ ActiveAdmin.register Fmu do
     column :certification_ls
   end
 
+  show do
+    attributes_table do
+      row :id
+      row :name
+      row :forest_type
+      row :country
+      row :operator
+      row :certification_fsc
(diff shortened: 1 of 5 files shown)


app/admin/fmu.rb
```diff
@@ -70,6 +70,28 @@ def scoped_collection
     column :certification_ls
   end
 
+  show do
```
tsubik, line 73: I wanted to have proper geojson to copy paste from.

santostiago (approved):
@tsubik all good. 
Just one question: is the frontend already sending the parameter `operator_id`?

tsubik:
Yes, I added that parameter to frontend for operator FMUs details map.

PR #309, opened 2022-06-24
Fix/update deployment tasks

tsubik:
I refactored deployment scripts for otp portal and observations tool. Here I'm refactoring and actually fixing deployments from the backoffice + small performance improvements.


santostiago (approved):
I just didn't understand why don't we refresh and redeploy from transifex (that's kind of the purpose of the buttons).
The rest is fine! :)

tsubik:
Everything is in `deploy` script take a look at otp-portal and tool repo and look into deploy script there.

PR #310, opened 2022-06-29
Fix application mailer

tsubik:
Fixing Application Mailer to send emails. I will do proper refactor and test suite in another task, this is just a hot fix.


santostiago (approved):
👍

PR #311, opened 2022-06-30
fix sentry not logging API issues

tsubik:


config/initializers/sentry.rb
```diff
@@ -7,7 +7,7 @@
   # Set tracesSampleRate to 1.0 to capture 100%
   # of transactions for performance monitoring.
   # We recommend adjusting this value in production
-  config.traces_sample_rate = 1.0
+  config.traces_sample_rate = 0.5
```
santostiago, line 10: I have it set to 0 in production in other projects. It's probably safe to turn it off.
But the sample rate should be `config.sample_rate = 1`. Is it `1` by default?

santostiago (approved):

PR #323, opened 2022-07-31
feature(Notifications): Add notification system

santostiago:
Adds a system that allows creating notifications groups and that generates daily notifications for documents that are about to expire.

This PR is just the backend + backoffice.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/admin/custom_admin_header.rb
+++ b/app/admin/custom_admin_header.rb
@@ -119,6 +119,14 @@ class CustomAdminHeader < ActiveAdmin::Views::Header
               li { link_to 'Users',              admin_users_path }
               li { link_to 'Access Control',     admin_access_control_path }
               li { link_to 'Comments',           admin_comments_path }
+              li do
+                text_node content_tag 'a', 'Notifications', class: '-with-children'
+                ul do
+                  li { link_to 'Notification Groups', admin_notification_groups_path }
+                  li { link_to 'Notifications', admin_notifications_path }
+                end
+              end
+
             end
           end
         end
--- a/app/admin/custom_admin_header.rb
+++ b/app/admin/custom_admin_header.rb
@@ -3,7 +3,8 @@
 class CustomAdminHeader < ActiveAdmin::Views::Header
   include Rails.application.routes.url_helpers
 
-  def build(namespace, menu)
+
+  def build(namespace, menu) # rubocop:disable Metrics/AbcSize
     div class: 'c-nav' do
       div class: 'logo' do
         div do
--- a/app/admin/notification.rb
+++ b/app/admin/notification.rb
@@ -23,9 +23,14 @@ ActiveAdmin.register Notification do
 
   actions :all, except: [:new, :edit, :update]
 
-  def scoped_collection
-    super.includes(operator_document: [:operator, :required_operator_document])
+  controller do
+    def scoped_collection
+      end_of_association_chain.includes(:user, :notification_group,
+                                        operator_document: [operator: [:translations],
+                                                            required_operator_document: [:translations]])
+    end
   end
+
 
   index do
     selectable_column
--- a/app/admin/notification.rb
+++ b/app/admin/notification.rb
@@ -11,15 +11,32 @@ ActiveAdmin.register Notification do
   scope :dismissed
   scope :solved
 
+  filter :notification_group
+  filter :operator_document_operator_id,
+         as: :select,
+         label: 'Operator',
+         collection: -> { Operator.with_translations(I18n.locale).order('operator_translations.name') }
+  filter :user
+  filter :last_displayed_at
+  filter :dismissed_at
+  filter :solved_at
+
   actions :all, except: [:new, :edit, :update]
+
+  def scoped_collection
+    super.includes(operator_document: [:operator, :required_operator_document])
+  end
 
   index do
     selectable_column
     column :user
-    column :operator
+    column :operator do |resource|
+      link_to(resource.operator_document.operator.name,
+              admin_producer_path(resource.operator_document.operator))
+    end
     column :operator_document do |resource|
       link_to(resource.operator_document.required_operator_document.name,
-              admin_operator_document_path(resource.operator_document.id))
+              admin_operator_document_path(resource.operator_document))
     end
     column :notification_group
     column :last_displayed_at
--- a/app/admin/notification.rb
+++ b/app/admin/notification.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register Notification do
+  extend BackRedirectable
+  back_redirect
+
+  menu false
+
+  scope :newly_created
+  scope :seen
+  scope :dismissed
+  scope :solved
+
+  actions :all, except: [:new, :edit, :update]
+
+  index do
+    selectable_column
+    column :user
+    column :operator
+    column :operator_document do |resource|
+      link_to(resource.operator_document.required_operator_document.name,
+              admin_operator_document_path(resource.operator_document.id))
+    end
+    column :notification_group
+    column :last_displayed_at
+    column :dismissed_at
+    column :solved_at
+    column :expiration_date do |resource|
+      resource.operator_document.expire_date
+    end
+    column :on_time? do |resource|
+      expire_date = resource.operator_document.expire_date
+      resource.solved_at || expire_date.blank? || expire_date > Date.today
+    end
+    column :created_at
+    column :updated_at
+    actions
+  end
+end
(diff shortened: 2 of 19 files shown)


app/models/operator_document.rb
```diff
@@ -149,6 +157,17 @@ def set_type
     self.type = 'OperatorDocumentCountry' if required_operator_document.is_a?(RequiredOperatorDocumentCountry)
   end
 
+  # Removes the existing notifications for an operator document
+  # as long as the new expire date is longer than the notification period or the group has been removed
+  def remove_notifications
+    Notification.left_joins(:notification_group).unsolved.where(operator_document_id: id).each do |notification|
```
tsubik, line 163: I think you meant `includes(:notification_group)` here, as you are not using notification_group in the query, and it is not sideloaded and probably should be as used later.

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
...
+from
+	operator_documents
+inner join required_operator_documents on
+	required_operator_documents.id = operator_documents.required_operator_document_id
+left outer join notifications on
+	notifications.operator_document_id = operator_documents.id and notifications.notification_group_id in (#{group_ids}) and notifications.solved_at is null
+where
+	operator_documents.deleted_at is null
+	and (expire_date < '#{date}'::date
+		and status = 3
```
tsubik, line 30: I would change that to `and status = #{OperatorDocument.statuses[:doc_valid]}` it would be more clear.

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
...
+      OperatorDocument.find_by_sql(sql(notification_date, group_ids.join(', '))).each do |od|
+        od.operator.users.each do |user|
+          Notification.create!(notification_group: notification_group, operator_document: od, user: user)
+        end
+      end
+    end
+  end
+
+  def sql(date, group_ids)
+    #TODO: This would be better in ARel
```
tsubik, line 17: I think it's ok to keep it as SQL, more readable haha

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
+namespace :notifications do
+  desc 'creates the notifications daily'
+  task create: :environment do
+    group_ids = []
+    NotificationGroup.order(days: :asc).each do |notification_group|
+      group_ids << notification_group.id
+      notification_date = Date.today + notification_group.days
+      OperatorDocument.find_by_sql(sql(notification_date, group_ids.join(', '))).each do |od|
```
tsubik, line 8: maybe worth to add `includes(operator: :users)` for better performance?

tsubik: and not sure if distinct or uniq is also needed

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
...
+    #TODO: This would be better in ARel
+    <<~SQL
+select
+	operator_documents.*
+from
+	operator_documents
+inner join required_operator_documents on
+	required_operator_documents.id = operator_documents.required_operator_document_id
+left outer join notifications on
+	notifications.operator_document_id = operator_documents.id and notifications.notification_group_id in (#{group_ids}) and notifications.solved_at is null
```
tsubik, line 26: hmm, I think this will keep adding notifications if someone `solve` them before expire_date?

spec/models/notification_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Notification, type: :model do
```
tsubik, line 3: Could test required associations presence.

spec/lib/tasks/create_notifications_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  let(:country) { create :country }
+  let(:required_operator_document) { create :required_operator_document_country, country: country }
+  let(:operator) { create :operator, country: country }
+  let!(:user1) { create :operator_user, country: country, operator: operator }
+  let!(:user2) { create :operator_user, country: country, operator: operator }
+  subject { Rake::Task["notifications:create"].invoke }
+
+  shared_examples 'no notifications change' do
```
tsubik, line 17: never used that in this way (same file), but actually good idea to DRY specs a little bit.

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
+namespace :notifications do
+  desc 'creates the notifications daily'
```
tsubik, line 2: We have all for cron in schuduler namespace, better to move it there

lib/tasks/create_notifications.rake
tsubik: Ok, solved is something else, so it should add new, then it's ok.

app/models/notification.rb
```diff
@@ -0,0 +1,39 @@
...
+
+# == Schema Information
+#
+# Table name: notifications
+#
+#  id                    :integer          not null, primary key
+#  last_displayed_at     :datetime
+#  dismissed_at          :datetime
+#  solved_at             :datetime
+#  operator_document_id  :integer
```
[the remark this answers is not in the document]
tsubik: I think none of them is optional, so all should be not null

lib/tasks/create_notifications.rake
```diff
@@ -0,0 +1,37 @@
...
+  task create: :environment do
+    group_ids = []
+    NotificationGroup.order(days: :asc).each do |notification_group|
+      group_ids << notification_group.id
+      notification_date = Date.today + notification_group.days
+      OperatorDocument.find_by_sql(sql(notification_date, group_ids.join(', '))).each do |od|
+        od.operator.users.each do |user|
+          Notification.create!(notification_group: notification_group, operator_document: od, user: user)
```
tsubik, line 10: `operator` is missing here?

app/admin/custom_admin_header.rb
```diff
@@ -3,7 +3,8 @@
 class CustomAdminHeader < ActiveAdmin::Views::Header
   include Rails.application.routes.url_helpers
 
-  def build(namespace, menu)
+
+  def build(namespace, menu) # rubocop:disable Metrics/AbcSize
```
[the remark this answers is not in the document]
santostiago: @tsubik [no need](https://stackoverflow.com/questions/37921807/if-a-rubocop-rule-is-disabled-in-line-does-it-need-to-be-re-enabled). :)

app/models/notification.rb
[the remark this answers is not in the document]
santostiago: @tsubik right... I thought the default was not null for Rails 5. I'll update it.

spec/models/notification_spec.rb
santostiago: @tsubik you mean to test that the association is mandatory? I share [Jason's opinion](https://www.codewithjason.com/examples-pointless-rspec-tests/) on this matter :P

lib/tasks/create_notifications.rake
santostiago: I moved it to the namespace... or do you mean the file?

lib/tasks/create_notifications.rake
santostiago: @tsubik yeah... I added `operator` just to be easier to see the notifications of an operator.. but to be fair, it gets denormalized and there's no real benefit to doing it... I can easily be fetched with a join... Maybe I should just remove it.

app/admin/notification.rb
```diff
@@ -0,0 +1,56 @@
...
+         label: 'Operator',
+         collection: -> { Operator.with_translations(I18n.locale).order('operator_translations.name') }
+  filter :user
+  filter :last_displayed_at
+  filter :dismissed_at
+  filter :solved_at
+
+  actions :all, except: [:new, :edit, :update]
+
+  def scoped_collection
```
santostiago, line 26: @tsubik this should fix the N+1 problem in AA, but it doesn't look like it does.

tsubik: you forgot `controller do` look at other admin files, not sure if that should be `super` or `end_of_association_chain`. I was always using `super` but this is old AA version as well, so could work differently.

lib/tasks/create_notifications.rake
tsubik: I meant the file, but maybe that is ok as well.

lib/tasks/create_notifications.rake
tsubik: So, if it's not used for anything you can remove it.

spec/models/notification_spec.rb
tsubik: Not that kind, or maybe similar, I meant presence validation. If we add not nulls for relations.

tsubik: Never seen `have_one` expectations, that must be some additional gem.

tsubik: At least that default `build(:notification)` is valid should be here.

tsubik (approved):
:+1: Great

PR #330, opened 2022-08-09
Fix moving attachments if that is blank

tsubik:
Connected to this error https://sentry.io/organizations/wri-otp/issues/3490029762


santostiago (approved):
👍

PR #332, opened 2022-08-17
Feature/obs tool additional filters

tsubik:
Adding extra filters for couple resources to get filtered results for observation tool.

- FMUs - return only ones that are linked to an observations from the IM
- Gov entities - return only ones that are linked to an observation from the IM
- Producers - return only ones that are linked to an observation from the IM


santostiago (approved):
👍

PR #333, opened 2022-08-17
Chore/improve test coverage

tsubik:
Little refactoring. Adding more specs, removing not needed files, vol1.


app/controllers/v1/observation_filters_controller.rb
```diff
@@ -255,7 +255,7 @@ def valid_params(name, value)
     # TODO: Hack. Modify to use refinements
     def self.is_of_type(type, value)
       if type == Integer
-        return true if value.to_i.to_s == self
+        return true if value.to_i.to_s == value
```
santostiago, line 258: 😮

app/models/concerns/globalize_fallback.rb
```diff
@@ -1,43 +0,0 @@
-# frozen_string_literal: true
```
santostiago, line 1: Mmmm... so how do we do now when there's a field missing in translations?

santostiago (approved):
I just have a question regarding the globalize fallback.

app/models/concerns/globalize_fallback.rb
tsubik: @santostiago that module is not used anywhere, at least I did not find any reference to it. Code coverage of this code was also 0%

PR #337, opened 2022-09-25
feature(Newsletter): Send quarterly newsletters to users

santostiago:
Adds a job that sends quarterly newsletters to the users of a producer with information about the changing of scores and documents to expire.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No email

Issue Number: https://www.pivotaltracker.com/story/show/181285357

## What is the new behavior?
With email!!! Yey!!

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/services/mail_service.rb
+++ b/app/services/mail_service.rb
@@ -197,6 +197,58 @@ TXT
     self
   end
 
+  # An email that contains the a quarterly report of an operator
+  # It lists:
+  # 1. Current transparency score
+  # 2. Change of score in the last quarter
+  # 3. List of documents expiring in the next quarter
+  # It's sent every quarter to all users of an operator
+  #
+  # TODO: Currently we don't have a way to know the language of the operator, so we send it in both French and English
+  # In the future this should be refactored:
+  # 1. To send it in the language of the user and use I18n
+  # 2. To send it using the html template
+  def quarterly_newsletter(operator)
+    current_score = operator.score_operator_document
+    last_score = operator.score_operator_documents.at_date(Date.today - 3.months).order(:date).last
+    expiring_docs = operator.operator_documents.to_expire(Date.today + 3.months)
+
+    subject = "Your quarterly report. / Votre rapport trimestriel."
+
+    current_score_percentage = NumberHelper.float_to_percentage(current_score.all)
+
+    text_en = ["Your current score is #{current_score_percentage}."]
+    text_fr = ["Votre score actuel est de #{current_score_percentage}."]
+
+    if last_score.present?
+      last_score_percentage = NumberHelper.float_to_percentage(last_score.all)
+
+      score_change = NumberHelper.float_to_percentage(current_score.all - last_score.all)
+      text_en << "Your score on #{last_score.date} was #{last_score_percentage}. This means a variation of #{score_change}."
+      text_fr << "Votre dernier score en #{last_score.date} était de #{last_score_percentage}. Cela signifie une variation de #{score_change}."
+    end
+
+    expiring_docs.each do |document|
+      text_en << "Document #{document.required_operator_document.name} expires on #{document.expire_date}."
+      text_fr << "Le document #{document.required_operator_document.name} expire en #{document.expire_date}."
+    end
+
+    text_en << ['', 'Best,', 'OTP Team', '']
+    text_fr << ['', 'Cordialement,', "L'équipe OTP", '']
+
+    body = text_en.join("\n")
+    body << "\n----------------------------------------------------\n"
+    body << text_fr.join("\n")
+
+    @from = ENV['CONTACT_EMAIL']
+    @to = operator.users.pluck(:email).join(', ')
+    @body = body
+    @subject = subject
+    @content_type = "text/html"
+
+    self
+  end
+
   private
 
   def document_admin_url(document)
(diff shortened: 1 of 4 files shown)


app/services/mail_service.rb
```diff
@@ -197,6 +197,54 @@ def notify_operator_expiring_document(operator, documents)
...
+  # 3. List of documents expiring in the next quarter
+  # It's sent every quarter to all users of an operator
+  #
+  # TODO: Currently we don't have a way to know the language of the operator, so we send it in both French and English
+  # In the future this should be refactored:
+  # 1. To send it in the language of the user and use I18n
+  # 2. To send it using the html template
+  def quarterly_newsletter(operator)
+    current_score = operator.score_operator_document
+    last_score = operator.score_operator_documents.at_date(Date.today - 3.months).last
```
tsubik, line 213: `order(:date)` as order is never guaranteed, better to make sure it's good.

app/services/mail_service.rb
```diff
@@ -197,6 +197,54 @@ def notify_operator_expiring_document(operator, documents)
...
+    expiring_docs = operator.operator_documents.to_expire(Date.today + 3.months)
+
+    subject = "Your quarterly report. / Votre rapport trimestriel."
+
+    text_en = ["Your current score is #{current_score.all}."]
```
tsubik, line 218: As score is a percentage of valid documents provided and I'm not sure if anyway is accustomed to its raw full value. We need to make sure scores are properly formatted as percentage, with up to 2 digits after decimal points. What do you think? Is the text already approved by Sophie?

lib/tasks/scheduler.rake
```diff
@@ -53,7 +53,16 @@ namespace :scheduler do
...
-    Rails.logger.info "Observation resports statistics generated. It took #{time} ms."
+    Rails.logger.info "Observation reports statistics generated. It took #{time} ms."
+    Rails.logger.info '::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::'
+  end
+
+  desc 'Send quarterly newsletters to operators'
+  task send_quarterly_newsletters: :environment do
+    Rails.logger.info '::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::'
+    Rails.logger.info "Going to send quarterly newsletters: #{Time.now.strftime('%d/%m/%Y %H:%M')}"
+    time = Benchmark.ms {  Operator.active.fa_operator.find_each { MailService.new.quarterly_newsletter(_1).deliver } }
```
tsubik, line 64: First time I see `_1` in usage and TBH I have mixed feelings about it, maybe because first time I see it.

tsubik: There could be a problem with sending emails in this way. We have to migrate to sidekiq or any other background job and send each email in separate job as here if anything goes wrong with one operator then no emails will be sent to the rest. You could add a rescue here and keep sending to the last email and after that fail the whole task to inform us that something was wrong. That would be better but not ideal in my opinion.

app/services/mail_service.rb
santostiago: @tsubik right!
It will look better that way.

lib/tasks/scheduler.rake
santostiago: @tsubik `_1` is the "recommended" way to use it now when the name of the block parameter is irrelevant. But probably here I can use `with_self` or similar.
True. It's definitely better to rescue it in case something goes wrong the other emails get sent (and we get logs letting us know something is wrong). Another thing worth checking is a service for error monitoring: we didn't add any, right?

tsubik: Ok. I added service to monitoring cron jobs (healthchecks.io) and sentry is for error monitoring as well.

tsubik (approved):
Approving, but still, I would add rescue in the block in the rake task when sending messages and manually report to sentry and fail the job after the rest is sent (or not sent).

santostiago:
@tsubik I haven't updated the PR to notify Sentry yet. I'll do it and re-request the review.

santostiago:
@tsubik I just updated it to notify sentry on error and proceed with the remaining operators.

tsubik (changes requested):
Added a comment in rake task

app/services/mail_service.rb
```diff
@@ -197,6 +197,58 @@ def notify_operator_expiring_document(operator, documents)
...
+  def quarterly_newsletter(operator)
+    current_score = operator.score_operator_document
+    last_score = operator.score_operator_documents.at_date(Date.today - 3.months).order(:date).last
+    expiring_docs = operator.operator_documents.to_expire(Date.today + 3.months)
+
+    subject = "Your quarterly report. / Votre rapport trimestriel."
+
+    current_score_percentage = NumberHelper.float_to_percentage(current_score.all)
+
+    text_en = ["Your current score is #{current_score_percentage.all}."]
```
tsubik, line 220: here and in the next line is `.all` to be removed.

santostiago: ups.

lib/tasks/scheduler.rake
```diff
@@ -59,7 +59,24 @@ namespace :scheduler do
...
+    time = Benchmark.ms do
+      Operator.active.fa_operator.find_each do |operator|
+        begin
+          MailService.new.quarterly_newsletter(operator).deliver
+        rescue StandardError => e
+          Sentry.capture_exception(e, extra: { 'operator_id' => operator.id })
+        end
+      end
+    end
+    Rails.logger.info "Sent quarterly newsletters to operators. It took #{time} ms."
```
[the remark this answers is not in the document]
tsubik: What do you think?

tsubik (approved):
I left a question if you think it would be beneficial to fail rake task at the end if there is any error, but other than that all good :+1:

santostiago:
@tsubik do we get any particular notification if the rake task fails?
Isn't it enough to notify Sentry for each error?

tsubik:
Yes we do have healtchecks for cronjobs, better to add it, if Sentry somehow fails then we will have a fail in other place.

PR #339, opened 2022-09-27
Chore/rails 51

tsubik:
First step to upgrade backend to newer Rails version.


app/controllers/v1/users_controller.rb
```diff
@@ -24,7 +24,7 @@ def update
     end
 
     def current
-      user = User.find(context[:current_user])
+      user = User.find(context[:current_user].id)
```
santostiago, line 27: mmm... this worked before? :O

app/controllers/v1/users_controller.rb
```diff
@@ -62,7 +62,7 @@ def update_params
           :locale,
         )
       # better in this way to keep actioncontroller::parameters object, this rails version has some bugs
-      p.deep_transform_keys!(&:underscore)
+      p.transform_keys!(&:underscore)
```
santostiago, line 65: why the change?

app/models/observation.rb
```diff
@@ -83,13 +83,13 @@ class Observation < ApplicationRecord
   attr_accessor :user_type
 
   belongs_to :country,        inverse_of: :observations
```
santostiago, line 85: shouldn't we also add the `optional: true` to the country? I'm not sure if it's mandatory in the observations tool.

app/models/operator_document_history.rb
```diff
@@ -31,11 +31,11 @@
...
 
-  belongs_to :operator, optional: false
-  belongs_to :required_operator_document, -> { with_archived }, required: true
+  belongs_to :operator
+  belongs_to :required_operator_document, -> { with_archived }
   belongs_to :fmu , -> { with_deleted },  optional: true
   belongs_to :user, optional: true
-  belongs_to :document_file, optional: :true
+  belongs_to :document_file, optional: true
   belongs_to :operator_document, -> { with_deleted }
```
santostiago, line 39: mmm... I have a feeling this should be optional. 
Isn't there a case in which the operator document can be removed, but the history remains? (maybe not...)

app/models/photo.rb
```diff
@@ -18,7 +18,7 @@
 class Photo < ApplicationRecord
   mount_base64_uploader :attachment, PhotoUploader
 
-  belongs_to :user, inverse_of: :photos
+  belongs_to :user, inverse_of: :photos, optional: true # TODO: I think this is model is not used could be removed
```
santostiago, line 21: Me too. Did you check the server to see if there are any files there?

spec/models/operator_document_spec.rb
```diff
@@ -188,14 +189,15 @@
       end
     end
 
-    describe '#insure_unity' do
+    describe '#ensure_unity' do
```
santostiago, line 192: I made this mistake billions of times....

spec/factories/operator_document_histories.rb
```diff
@@ -28,13 +28,14 @@
 #
 
 FactoryBot.define do
-  factory :operator_document_history do
+  factory :operator_document_history, class: OperatorDocumentCountryHistory, aliases: [:operator_document_country_history] do
```
santostiago, line 31: I just didn't understand this part (and the similar ones below)

santostiago (approved):

app/controllers/v1/users_controller.rb
tsubik: Yes, in 5.0 it worked, although I've never seen that method used in that way, first time here. In 5.1 that's ArgumentError telling to use id.

app/controllers/v1/users_controller.rb
tsubik: ActiveController::Parameters has no longer all Hash object methods, so there is no `deep_transform_keys` let me dig into history why key transform was added here. Comment is interesting saying there was some bug in the framework, maybe it's gone in 5.1 version, so I could just remove that method.

app/models/observation.rb
tsubik: It's ok, no nulls in the database, so I guess it's required or there are no cases to be optional.

app/models/operator_document_history.rb
tsubik: As document is soft-deleted then all those cases are bad data cases as it does not make sense to operator_document to be optional. We don't have any foreign keys in db, so one of the next PRs will add them and we have to deal with bad data.

app/models/photo.rb
tsubik: Nothing in the database and nothing on the disk. I will remove that model.

spec/factories/operator_document_histories.rb
tsubik: Well, if possible better to stay away from STI as it has many shortcomings. Here there was some problem with factories if I have not set class I think. By default, I made `operator_document_history` factory to create country document. Maybe there should be no `operator_document_history` factory at all, only country and fmu ones, but I did not want to change all the specs.

app/controllers/v1/users_controller.rb
tsubik: ok, I need that `transform_keys` here, but it does not have to be deep, so all good.

app/models/photo.rb
tsubik: but in the next PR

tsubik: to keep them more tidy.

PR #340, opened 2022-09-29
Fix/creating fmu documents

tsubik:
This PR solves the issue with mismatched documents created for FMU (the code was not looking at fmu forest types at all in create_documents hook on operator update)


santostiago (approved):
We should also remove the calls to `rebuild_documents` from `migrations.rake` and `migrations-v2.rake`.

tsubik:
Ahh, yes. I wanted to ask about those files. I would probably just remove those files @santostiago , but would be also great to clean up rake tasks at some point.

PR #344, opened 2022-10-18
Change sorting column in admin panel to prevent server crash

tsubik:
I think that this is because the old version of active admin which uses old version of ransack. When sorting by for example countries_translations.name in operator LIMIT and OFFSET is removed from query what is causing the whole dataset to load.


santostiago (approved):

santostiago:
I'll also check the timeouts for AR and the DB

tsubik:
I think that might be something with Postgres (like using old version), because it just takes all CPU for those queries after couple seconds and system freezes.

PR #346, opened 2022-10-19
fix: update score when documents deleted

tsubik:


app/decorators/score_operator_document_decorator.rb
```diff
@@ -41,8 +41,14 @@ def print_diff(prev, current)
       elsif value < prev[key]
         color = 'red'
       end
-
-      "#{key.gsub('doc_','')}: <span style='color: #{color}'>#{value}</span>"
+      diff = value - prev[key]
```
tsubik, line 44: This fragment is added for quicker debug using admin panel.

app/models/fmu_operator.rb
```diff
@@ -99,14 +99,19 @@ def self.calculate_current
...
     current_operator = self&.fmu&.reload&.operator
 
     OperatorDocumentFmu.transaction do
-      to_destroy = OperatorDocumentFmu.where(fmu_id: fmu_id).where.not(operator_id: current_operator&.id)
+      to_destroy = OperatorDocumentFmu.includes(:operator).where(fmu_id: fmu_id)
+      to_destroy = to_destroy.where.not(operator_id: current_operator.id) if current_operator.present?
       destroyed_count = to_destroy.count
-      to_destroy.each { |x| x.destroy }
+      to_destroy.each do |doc|
+        doc.skip_score_recalculation = true # just do it once for each operator at the end
```
santostiago, line 111: yeah, it's better that way.

app/models/required_operator_document.rb
```diff
@@ -45,26 +45,18 @@ class RequiredOperatorDocument < ApplicationRecord
 
   validate :fixed_fields_unchanged
 
-  before_destroy :invalidate_operator_documents, prepend: true
   after_restore :set_documents_not_provided
 
   scope :with_archived, -> { unscope(where: :deleted_at) }
 
-  def invalidate_operator_documents
```
santostiago, line 53: We have `dependent: :destroy` for operator documents. 
I can't remember if paranoia destroys them or if it performs soft delete.
If it performs soft delete, we should leave this code so that the documents go expired. WDYT?

lib/tasks/sync.rake
```diff
@@ -103,6 +103,25 @@ class SyncTasks
       end
     end
 
+    operators = Operator.all
```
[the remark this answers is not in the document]
santostiago: When and why do we run these jobs?

spec/models/operator_document_spec.rb
```diff
@@ -147,6 +147,38 @@
       end
     end
 
+    describe '#recalculate operator score' do
+      before { @document = create(:operator_document_fmu, fmu: @fmu, operator: @operator) }
```
santostiago, line 151: why are these 2 sentences not inside the same `before` block?

spec/models/operator_document_spec.rb
```diff
@@ -147,6 +147,38 @@
...
+      before { expect(ScoreOperatorDocument).to receive(:recalculate!).with(@operator) }
+
+      context 'when removing document' do
+        it 'updates operator score' do
```
santostiago, line 155: mmm.. this one is not very clear. What's the expectation?

santostiago (approved):
Left just a few notes and questions.

lib/tasks/sync.rake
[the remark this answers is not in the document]
tsubik: Just occasionally. I left that job to be able to fix scores and scores history when there is something wrong with the data. I was wondering if that not should be done daily, but in that case we may not notice if something is wrong with the system that is causing scores between frontend and backend to go out of sync.

app/models/required_operator_document.rb
tsubik: It performs soft-delete which is what we actually want, no? The comment that was in `regenerate` method says it should soft delete when this is removed, but here we had a code that made documents expired. Those two are contradictory, which one is correct? Maybe we can double check with Sophie

spec/models/operator_document_spec.rb
tsubik: I don't think that matters, but I can change it.

spec/models/operator_document_spec.rb
tsubik: Hmmm, why is not clear? When you remove the document, then operator score is updated.

tsubik: This is what was not happening that's why we had scores on frontend going out of sync.

PR #348, opened 2022-10-21
Remove photo model

tsubik:


santostiago (approved):
All good. I noticed we don't have any photos in production. 
Isn't there anything in the observations tool to upload them? They only upload evidences for the observations?

tsubik:
Yes, only evidences (observation document model) are uploaded and it could be anything (document, photo, anything). There are some models for photo API but those are not used so I will also remove them from observation tool.

PR #350, opened 2022-10-26
Feature/obs gps coordinates

tsubik:
Feature already tested and approved https://www.pivotaltracker.com/story/show/181285725

### DIFF
--- a/app/views/admin/observations/_map.html.erb
+++ b/app/views/admin/observations/_map.html.erb
@@ -26,6 +26,13 @@
 
   map.on('load', function() {
     <% if bbox.present? %>
+    map.once('moveend', () => {
+      setTimeout(() => {
+        const newBounds = map.getBounds();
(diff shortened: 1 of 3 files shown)


app/views/admin/observations/_map.html.erb
```diff
@@ -0,0 +1,70 @@
+<script src="https://api.mapbox.com/mapbox-gl-js/v1.9.0/mapbox-gl.js"></script>
```
santostiago, line 1: Didn't we have one for the FMU already? Was it no reusable?

santostiago (approved):

app/views/admin/observations/_map.html.erb
tsubik: I copied just the fragment from that code, as making it reusable would be too much hassle.

PR #351, opened 2022-10-27
Low hanging fruit to improve API performance a bit

tsubik:


santostiago (approved):
This saves a few queries, right?

tsubik:
Overall gain is noticeable.

PR #354, opened 2022-10-28
Feature/return hidden observations api

tsubik:
- changing hidden filter for observation API
- remove not used observations CSV export

### DIFF
--- a/app/controllers/v1/observation_filters_controller.rb
+++ b/app/controllers/v1/observation_filters_controller.rb
@@ -6,39 +6,11 @@ module V1
 
     skip_before_action :authenticate
 
-    FILTER_TYPES = {
-      'validation-status': { type: /Published \(no comments\)|Published \(not modified\)|Published \(modified\)/ },
-      'observation-type': { type: /operator|government/ },
-      'country-id': { type: Integer },
-      'fmu-id': { type: Integer },
-      'year': { type: Integer, query: 'EXTRACT(year FROM observations.publication_date) IN ' },
-      'observer-id': { type: Integer, query: 'observers.id IN ' },
-      'category-id': { type: Integer, query: 'categories.id IN ' },
-      'subcategory-id': { type: Integer },
-      'severity-level': { type: Integer, query: 'severities.level IN ' },
-      'operator': { type: Integer },
-      'observation-report.id': { type: Integer, query: 'observation_reports.id IN ' },
-      'source': { type: Integer },
-      'hidden': { type: /true|false|all/ }
-    }.freeze
-
-    OBS_TYPES = {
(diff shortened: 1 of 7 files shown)


app/models/country.rb
```diff
@@ -54,13 +54,9 @@ class Country < ApplicationRecord
...
 
-  scope :with_observations, (-> {
-    left_outer_joins(:observations).where.not(observations: { id: nil }).distinct
-  })
-
-  scope :with_active_observations, (-> {
-    joins(:observations).merge(Observation.active).where.not(observations: { id: nil }).distinct
-  })
+  scope :with_observations, ->(scope = Observation.all) {
+    joins(:observations).merge(scope).where.not(observations: { id: nil }).distinct
```
santostiago, line 58: don't we want to keep the countries that have no observations?

app/resources/v1/observation_resource.rb
```diff
@@ -40,8 +40,10 @@ class ObservationResource < JSONAPI::Resource
             :observation_report, :law, :operator, :subcategory,
             :is_active, :validation_status, :is_physical_place
 
-    filter :hidden, apply: ->(records, value, _options) {
-      records.unscope(where: :hidden).where(hidden: value)
+    filter :hidden, default: 'false', apply: ->(records, value, options) {
+      return records if value.include?('all')
```
santostiago, line 44: will this work? Won't the default scope override it?

app/controllers/v1/observation_filters_controller.rb
```diff
@@ -69,113 +42,8 @@ def tree
       render json: filters
     end
 
-    def csv
```
santostiago, line 72: I can't remember where this was used. Did you check it?

santostiago (approved):

app/models/country.rb
tsubik: No, why? The scope is with_observations, I just changed the method to take with what kind of observations.

app/resources/v1/observation_resource.rb
tsubik: I removed default scope, and I'm using default filter (`false`). Anyway, I think the usage of `unscope` is some kind of code smell, or quick workaround and probably should be refactored.

app/controllers/v1/observation_filters_controller.rb
tsubik: I asked Sophie, she said to remove it.

tsubik: They are not using any CSV download, and this method has comment pleading for refactor, so the best refactor is to remove code if not needed :D

PR #358, opened 2022-11-06
feature(Documents): Expire non required documents

santostiago:
[Task](https://www.pivotaltracker.com/story/show/183467740).

This PR adds "not-required" documents to the list of documents to expire.

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Only valid documents get expired

Issue Number: 183467740

## What is the new behavior?
Not required documents also get expired.

- [ ] Yes
- [x] No

### DIFF
--- a/app/admin/required_gov_document.rb
+++ b/app/admin/required_gov_document.rb
@@ -20,12 +20,11 @@ ActiveAdmin.register RequiredGovDocument do
       rd.deleted_at.nil?
     end
     column 'required_gov_document_group' do |rd|
-      rd.required_operator_document_group&.name
+      rd.required_gov_document_group&.name
     end
     column 'country' do |rd|
       rd.country&.name
     end
-    column :type
     column :name
 
   end
--- a/app/models/operator_document.rb
+++ b/app/models/operator_document.rb
@@ -80,7 +80,10 @@ class OperatorDocument < ApplicationRecord
   joins(:required_operator_document).where(required_operator_documents: { contract_signature: false })
 }                                                  # non signature
   scope :to_expire,                              ->(date) {
-  joins(:required_operator_document).where("expire_date < '#{date}'::date and status = #{OperatorDocument.statuses[:doc_valid]} and required_operator_documents.contract_signature = false")
+  joins(:required_operator_document)
+    .where('expire_date < ?', date)
+    .where(status: EXPIRABLE_STATUSES)
+    .where(required_operator_documents: { contract_signature: false })
 }
 
   enum status: { doc_not_provided: 0, doc_pending: 1, doc_invalid: 2, doc_valid: 3, doc_expired: 4, doc_not_required: 5 }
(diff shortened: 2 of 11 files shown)


app/models/operator_document.rb
```diff
@@ -80,14 +80,18 @@ class OperatorDocument < ApplicationRecord
...
+    .where(status: EXPIRABLE_STATUSES)
+    .where(required_operator_documents: { contract_signature: false })
 }
 
   enum status: { doc_not_provided: 0, doc_pending: 1, doc_invalid: 2, doc_valid: 3, doc_expired: 4, doc_not_required: 5 }
   enum uploaded_by: { operator: 1, monitor: 2, admin: 3, other: 4 }
   enum source: { company: 1, forest_atlas: 2, other_source: 3 }
 
   NON_HISTORICAL_ATTRIBUTES = %w[id attachment updated_at created_at].freeze
+  EXPIRABLE_STATUSES = statuses.slice(:doc_valid, :doc_expired).values.freeze
```
tsubik, line 94: `%w[doc_valid doc_expired]` or `%i[doc_valid doc_expired]` would be enough and it should work I think. Rails itself should change those values to integer.

tsubik (commented):
You can add another spec, pretty simple for `#expire_documents` in `operator_document_spec.rb` just to add another document with different status

tsubik (approved):
Other than missing spec, I think all good. Expiration task should that we already have running daily should expire those not required documents.

santostiago:
@tsubik I updated the PR to include tests.
I also realized we were not expiring gov documents, so I added them to the job (I'll confirm it with Sophie, but I think we weren't expiring them by mistake, and not by design).

santostiago:
Also, we should (later) refactor the `gov documents` and `operator documents` to include a `documents` module. Like 50% of the code is the same.

app/admin/required_gov_document.rb
```diff
@@ -20,12 +20,11 @@
       rd.deleted_at.nil?
     end
     column 'required_gov_document_group' do |rd|
-      rd.required_operator_document_group&.name
+      rd.required_gov_document_group&.name
```
tsubik, line 23: :+1:

spec/factories/gov_documents.rb
```diff
@@ -0,0 +1,21 @@
+FactoryBot.define do
+  factory :gov_document, class: GovDocument do
+    required_gov_document
+    expire_date { Date.tomorrow }
+    start_date { Date.yesterday }
+    current { true }
+
+    transient do
+      force_status { nil }
```
tsubik, line 9: Wouldn't work if we just set status? Anyway, interesting use of transient.

tsubik (approved):
Looks great. New specs are very well organized. I'm always struggling with that ;)

spec/factories/gov_documents.rb
santostiago: I always do it like this. I'm not sure if `status` would work.

PR #361, opened 2022-11-12
importer(Documents): Import Generic Operator Documents

santostiago:
[Task](https://www.pivotaltracker.com/story/show/181285797)

This commit adds a rake task to import the generic operator documents (documents that belong to no country and should be generated for particular operators that work in countries not included in OTP)

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

### DIFF
--- a/app/admin/operator_document.rb
+++ b/app/admin/operator_document.rb
@@ -248,10 +248,17 @@ ActiveAdmin.register OperatorDocument do
          collection: -> { Country.with_translations(I18n.locale).order('country_translations.name') }
   filter :required_operator_document,
          collection: -> {
-           RequiredOperatorDocument
-             .joins(country: :translations)
-             .order('required_operator_documents.name')
-             .where(country_translations: { locale: I18n.locale }).all.map { |x| ["#{x.name} - #{x.country.name}", x.id] }
+           rod_table = RequiredOperatorDocument.arel_table
+           country_t_table = Country::Translation.arel_table
+           country_name = Arel::Nodes::NamedFunction.new('coalesce', [country_t_table[:name], Arel::Nodes::SqlLiteral.new("'Generic'")]).as('country_name')
+
+           query =
+             RequiredOperatorDocument
+               .select(rod_table[:name], country_name, rod_table[:id])
+               .left_joins(country: :translations)
+               .on("country_translations.country_id = countries.id and country_translations.locale = '#{I18n.locale}'")
+               .order('required_operator_documents.name')
+           RequiredOperatorDocument.find_by_sql(query.to_sql).map { ["#{_1[:name]} - #{_1[:country_name]}", _1[:id]] }
          }
   filter :operator, label: 'Operator', as: :select,
                     collection: -> { Operator.with_translations(I18n.locale).order('operator_translations.name') }
--- a/app/models/required_operator_document_fmu.rb
+++ b/app/models/required_operator_document_fmu.rb
@@ -41,6 +41,8 @@ class RequiredOperatorDocumentFmu < RequiredOperatorDocument
   end
 
   def fmus
+    return Fmu.where.not(country_id: Country.active) if country_id.blank?
+
     fmu_attributes = { country_id: self.country_id }
     fmu_attributes[:forest_type] = self.forest_types if self.forest_types.present?
     Fmu.where(fmu_attributes)
--- a/lib/tasks/data_import.rake
+++ b/lib/tasks/data_import.rake
@@ -55,6 +55,26 @@ namespace :import do
   end
 
 
+  desc 'Import Generic Operator Documents'
+  task generic_operator_documents: :environment do
+    filename = File.expand_path(File.join(Rails.root, 'db', 'files', 'generic operator documents.csv'))
+    RequiredOperatorDocumentGroup.transaction do
+      CSV.foreach(filename, col_sep: ',', row_sep: :auto, headers: true, encoding: 'UTF-8').with_index do |row, i|
+        puts "Line #{i} - #{row}"
+
+        rodg = RequiredOperatorDocumentGroup.find_by!(name: row['Legal category'])
+        rod = row['Type'] == 'FMU' ? RequiredOperatorDocumentFmu : RequiredOperatorDocumentCountry
+
+        rod.create!(
+          required_operator_document_group: rodg,
+          valid_period: 365,
+          name: row['Generic document']
+        )
+      end
+    end
+  end
+
+
   desc 'Import Operator Documents Types'
   task operator_document_types: :environment do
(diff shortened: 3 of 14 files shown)


app/admin/operator_document.rb
```diff
@@ -251,7 +251,7 @@ def scoped_collection
            RequiredOperatorDocument
              .joins(country: :translations)
              .order('required_operator_documents.name')
-             .where(country_translations: { locale: I18n.locale }).all.map { |x| ["#{x.name} - #{x.country.name}", x.id] }
+             .where(country_translations: { locale: I18n.locale }).all.map { |x| ["#{x.name} - #{x.country&.name}", x.id] }
```
tsubik, line 254: maybe we should add `display_name` property to the model, as with generic document we either want to not have `-` or have ` - Generic` suffix.

app/admin/operator_documents_dashboard.rb
```diff
@@ -26,7 +26,7 @@
       if resource.country.nil?
         'All Countries'
       else
-        link_to resource.country.name, admin_country_path(resource.country)
+        link_to resource.country.name, admin_country_path(resource.country) if resource.country.present?
```
tsubik, line 29: This is not needed here as the `else` is already for `resource.country.present?`

app/admin/operator_document.rb
tsubik: Also that won't work as there is `inner join` in the scope and where on country_translations with specific locale. So it won't show generic documents in that list.

lib/tasks/data_import.rake
```diff
@@ -55,6 +55,24 @@ namespace :import do
...
 
 
+  desc 'Import Generic Operator Documents'
+  task generic_operator_documents: :environment do
+    filename = File.expand_path(File.join(Rails.root, 'db', 'files', 'generic operator documents.csv'))
+    RequiredOperatorDocumentGroup.transaction do
+      CSV.foreach(filename, col_sep: ',', row_sep: :auto, headers: true, encoding: 'UTF-8').with_index do |row, i|
+        puts "Line #{i}"
+
+        rodg = RequiredOperatorDocumentGroup.find_by!(name: row['Legal category'])
```
tsubik, line 65: Did you check it on production dump? It did not work, and I had to remove trailing spaces from 2 document categories. I did it on production and staging already.

tsubik: Documents were created only for handful (9 exactly) non active operators.

tsubik (changes requested):
Couple changes are necessary with filters and creation of documents, on which I guess the decision is still pending.

app/admin/operator_document.rb
santostiago: > Also that won't work as there is `inner join` in the scope and where on country_translations with specific locale. So it won't show generic documents in that list.

@tsubik I think we talked about this before: don't the translations get generated when the models are saved (if the translations were null)? I thought that was the behavior.

app/admin/operator_documents_dashboard.rb
santostiago: Ups... missed that one.

lib/tasks/data_import.rake
santostiago: @tsubik I'm going to check the generation of documents again. They shouldn't be generated for anyone.

app/admin/operator_document.rb
tsubik: Yes, but here is other problem. Now,  the new Required operator document won't have country and you are using `inner join` + where in that joined translation table, please change it to make sure new required operator document will also be on the list, but with better names. Could be just without country name and "-" sign.

app/admin/operator_documents_dashboard.rb
santostiago: I don't know why I did this..

lib/tasks/data_import.rake
santostiago: So, the rule is to do the same as for other ROD when they're added.

app/admin/operator_document.rb
santostiago: `display_name` on the model would require a query to fetch the country translation per each ROD.
I ended up doing it all in sql.

app/models/required_operator_document_fmu.rb
```diff
@@ -41,6 +41,8 @@ def create_operator_document_fmus
   end
 
   def fmus
+    return Fmu.where.not(country_id: Country.active.uniq.pluck(:id)) if country_id.blank?
```
tsubik, line 44: I think `not(country: Country.active)` would have the same result in generated AR queries (but I'm not sure) - plus is that would be more readable.

tsubik: but I don't understand this line.

tsubik: If `country_id` is blank then the document is for all non active countries fmus? Why?

tsubik: actually queries will be different, with `country: Country.active` there will be only one query.

lib/tasks/data_import.rake
tsubik: Works on the dump now, and no documents were created.

app/models/required_operator_document_fmu.rb
santostiago: > If country_id is blank then the document is for all non active countries fmus? Why?

@tsubik that's the rule.
When we create a generic ROD (whether they're for countries or FMUs), they are required for countries/fmus that are not in the list of active countries.
Active countries have a proper list of ROD. All the other ones have this generic list (which is shorter and "generic)

lib/tasks/data_import.rake
santostiago: @tsubik yes. There are no operators in other countries that have an `fa_id`.
This will have to be changed by Sophie.

app/models/required_operator_document_fmu.rb
tsubik: Ohh, ok, now that's clear.

app/models/required_operator_document_fmu.rb
```diff
@@ -41,6 +41,8 @@ def create_operator_document_fmus
   end
 
   def fmus
+    return Fmu.where.not(country_id: Country.active) if country_id.blank?
```
tsubik, line 44: Ok, now it's clear for me that generic are also for fmu documents for non active countries. So, `RequiredOperatorDocumentFmu` is generic if it does not have country_id, right? Maybe we can add method

```
def generic?
  country_id.blank?
end
```
and this code would be clear at first sight.
```

tsubik (approved):
:+1:

PR #362, opened 2022-11-13
Feat add positions to required documents

santostiago:
[Task](https://www.pivotaltracker.com/story/show/181285845)

Adds the field `position` to `RequiredOperatorDocument` and `RequiredGovDocument` that should be unique within the `Country` and the `RequiredDocumentGroup`.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

Issue Number: https://www.pivotaltracker.com/story/show/181285845

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/models/required_gov_document.rb
+++ b/app/models/required_gov_document.rb
@@ -22,6 +22,7 @@ class RequiredGovDocument < ApplicationRecord
   has_paper_trail
   includes Translatable
   acts_as_paranoid
+  acts_as_list scope: [:country_id, :required_gov_document_group_id, deleted_at: nil]
 
   translates :explanation, paranoia: true, touch: true, versioning: :paper_trail
   # rubocop:disable Style/BlockDelimiters
--- a/app/models/required_gov_document.rb
+++ b/app/models/required_gov_document.rb
@@ -13,7 +13,7 @@
 #  country_id                     :integer
 #  created_at                     :datetime         not null
 #  updated_at                     :datetime         not null
-#  required_gov_document_id       :integer          not null
+#  position                       :integer
 #  explanation                    :text
 #  deleted_at                     :datetime
 #
--- a/db/migrate/20221112185114_add_position_to_documents.rb
+++ b/db/migrate/20221112185114_add_position_to_documents.rb
@@ -1,0 +1,38 @@
+class AddPositionToDocuments < ActiveRecord::Migration[5.1]
+  def change
+    add_column :required_operator_documents, :position, :integer
+    add_column :required_gov_documents, :position, :integer
+
+    RequiredOperatorDocument.group(:country_id).pluck(:country_id).each do |country_id|
+      RequiredOperatorDocumentGroup.find_each do |rodg|
+        RequiredOperatorDocument
+          .with_translations
+          .where(country_id: country_id, required_operator_document_group: rodg)
+          .order(:name)
+          .each.with_index do |rod, index|
+          rod.update_column(:position, index + 1)
+        end
+      end
+
+    end
+
+    RequiredGovDocument.group(:country_id).pluck(:country_id).each do |country_id|
+      RequiredOperatorDocumentGroup.find_each do |rodg|
+        RequiredGovDocument
+          .with_translations
+          .where(country_id: country_id, required_operator_document_group: rodg)
+          .order(:name)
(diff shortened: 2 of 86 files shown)


app/models/required_gov_document.rb
```diff
@@ -37,6 +37,7 @@ class RequiredGovDocument < ApplicationRecord
   validates_presence_of :name
   validates_inclusion_of :document_type, in: RequiredGovDocument.document_types.keys
   validates :valid_period, numericality: { greater_than: 0 }, if: :valid_period?
+  validates :position, uniqueness: { scope: %w[country_id required_gov_document_group_id] }, allow_nil: true
```
tsubik, line 40: can we use `acts_as_list` instead? We are already using it in about page entries. I think I added that there to have better UX for the user (no need to manually add priority or changing priority will change the rest).

spec/models/required_operator_document_spec.rb
```diff
@@ -88,6 +88,63 @@
         end
       end
     end
+
+    describe '#validate_position' do
```
tsubik, line 92: no need for this if we use `acts_as_list`

tsubik (changes requested):
I think we should use acts_as_list

santostiago:
@tsubik `acts_as_list` is gonna be messy for these models.
Because it enforces uniqueness within a scope, when we delete required documents, there are going to be collisions of the positions and we'll have to hack around it, making it more complicated than it should.
WDYT?

santostiago:
@tsubik added `acts_as_list`.

db/migrate/20221112185114_add_position_to_documents.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    RequiredOperatorDocument.group(:country_id).pluck(:country_id).each do |country_id|
+      RequiredOperatorDocument.where(country_id: country_id).order(:created_at).each.with_index do |rod, index|
+        rod.update_column(:position, index)
+      end
+    end
+
+    RequiredGovDocument.group(:country_id).pluck(:country_id).each do |country_id|
+      RequiredGovDocument.where(country_id: country_id).order(:created_at).each.with_index do |rgd, index|
+        rgd.update_column(:position, index + 1)
```
tsubik, line 14: Ohh, not sure if that is a good idea. Right now on the frontend I sort by first position then name, as sorting by name was by default on the frontend. If you update position to be sorted by created_at date then the whole order will be different. You should order by name.

tsubik (changes requested):
@santostiago please add `position` attribute to jsonapi resources.

tsubik:
You can test this PR with current `develop` version of otp portal, or staging.

db/migrate/20221112185114_add_position_to_documents.rb
tsubik: Keeping position as `null` would also work on frontend, as if position is null then frontend sort the rest by title.

santostiago: @tsubik ok, I'll change it to be ordered by name.

santostiago: @tsubik and this is from `acts_as_list` documentation:
> As it stands acts_as_list requires position values to be set on the model before the instance methods above will work. Adding something like the below to your migration will set the default position. Change the parameters to order if you want a different initial ordering.

santostiago: (I mean, we don't really need those methods, but it's better to have everything ordered instead of requiring them to add a position for all)

tsubik:
I tried it and you don't have to set position, could be null and it worked
fine as well, but we can set it if it's not too hard.

santostiago:
@tsubik ready.
(I didn't understand the testing part)

tsubik:
I meant, frontend changes are already deployed, so you can test it locally with otp portal.

santostiago:
@tsubik added the sorting to consider the RODG.
Checked it on the BO, but I'll configure the portal tomorrow.

tsubik (approved):
:+1:

PR #363, opened 2022-11-13
Feat allow admin to edit observations

santostiago:
[Task](https://www.pivotaltracker.com/story/show/181285745)
When a flag to override observations is on, the admins can change observations in ways that are forbidden under normal circumstances.
The env var `OVERRIDE_OBSERVATION_STATUS` should be set to `TRUE`.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Admins cannot edit all the fields in observations

Issue Number: #181285745

## What is the new behavior?
Admins can now edit all the fields in observations

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/admin/observation.rb
+++ b/app/admin/observation.rb
@@ -23,6 +23,12 @@ ActiveAdmin.register Observation do
     end
   end
 
+  before_action do
+    if %w[POST PATCH].include?(request.method) && action_name != 'batch_action'
+      resource.user_type = :admin
+    end
+  end
+
   scope_to do
     Class.new do
       def self.observations
@@ -32,28 +38,42 @@ ActiveAdmin.register Observation do
   end
 
   actions :all, except: [:new]
-  permit_params :name, :lng, :pv, :lat, :lon, :subcategory_id, :severity_id, :operator_id,
-                :validation_status, :publication_date, :is_active, :observation_report_id,
-                :location_information, :evidence_type, :evidence_on_report, :location_accuracy,
-                :law_id, :fmu_id, :hidden, :admin_comment, :monitor_comment, :actions_taken,
-                :responsible_admin_id, observer_ids: [], relevant_operator_ids: [], government_ids: [],
-                                       observation_documents_attributes: [:id, :name, :attachment],
-                                       translations_attributes: [:id, :locale, :details, :concern_opinion, :litigation_status, :_destroy]
-
+  controller do
+    def permitted_params
+      if current_user.user_permission.user_role == 'bo_manager'
+        params.permit(observation: [:validation_status, :admin_comment])
+      else
+        params.permit(
+          observation: [
+            :name, :lng, :pv, :lat, :lon, :subcategory_id, :severity_id, :country_id, :operator_id, :user_type,
+            :validation_status, :publication_date, :observation_report_id, :location_information, :evidence_type,
+            :evidence_on_report, :location_accuracy, :law_id, :fmu_id, :hidden, :admin_comment,
+            :monitor_comment, :actions_taken, :responsible_admin_id, :is_physical_place,
+            observer_ids: [], relevant_operator_ids: [], government_ids: [],
+            observation_documents_attributes: [:id, :name, :attachment],
+            translations_attributes: [:id, :locale, :details, :concern_opinion, :litigation_status, :_destroy]
+          ]
+        )
+      end
+    end
+  end
 
   member_action :ready_for_publication, method: :put do
-    resource.update(validation_status: Observation.validation_statuses['Ready for publication'])
-    redirect_to collection_path, notice: 'Observation approved'
+    resource.validation_status = Observation.validation_statuses['Ready for publication']
+    notice = resource.save ?  'Observation moved to Ready for Publication' : 'Observation NOT modified'
+    redirect_to collection_path, notice: notice
   end
 
   member_action :needs_revision, method: :put do
-    resource.update(validation_status: Observation.validation_statuses['Needs revision'])
-    redirect_to collection_path, notice: 'Observation needs revision'
+    resource.validation_status = Observation.validation_statuses['Needs revision']
+    notice = resource.save ?  'Observation moved to Needs Revision' : 'Observation NOT modified'
+    redirect_to collection_path, notice: notice
   end
 
   member_action :start_qc, method: :put do
-    resource.update(validation_status: Observation.validation_statuses['QC in progress'])
-    redirect_to collection_path, notice: 'QC in progress for observations'
+    resource.validation_status = Observation.validation_statuses['QC in progress']
+    notice = resource.save ?  'Observation moved to QC in Progress' : 'Observation not modified'
+    redirect_to collection_path, notice: notice
   end
 
   action_item :ready_for_publication, only: :show do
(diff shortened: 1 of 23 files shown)


santostiago:
@tsubik I updated the `active admin addons` so that we could use nested fields.
This removes the ability to have image/file rows/columns in the admin.
Maybe I could check the code of the old gem to allow displaying images/file types, but it's not super important so I didn't add it yet.

santostiago:
Also, the field `law` is not dependent on the country/subcategory.
I could add some custom JS to make those ajax requests, but I thought it was too much trouble for the gains we'd get.

app/admin/observation.rb
```diff
@@ -431,15 +432,59 @@ def self.observations
...
-      f.input :subcategory, input_html: { disabled: true }
-      f.input :law, collection: Law.by_country_subcategory(object).map { |l| [l.written_infraction, l.id] } rescue ''
-      f.input :severity, as: :select,
-                         collection: object.subcategory.severities.map { |s| ["#{s.level} - #{s.details.first(80)}", s.id] } rescue ''
-      f.input :is_physical_place, input_html: { disabled: true }
+      if allow_override
+        if f.object.observation_type == 'operator'
+          f.input :fmu_id, as: :nested_select,
+                           level_1: {
+                    attribute: :country_id,
```
tsubik, line 440: indent level is weird here.

app/models/observation.rb
```diff
@@ -223,6 +224,12 @@ def published?
 
   private
 
+  # Returns whether the admin can override the status of an observations and its fields.
+  # This is defined by the ENV var `OVERRIDE_OBSERVATION_STATUS` which should be 'TRUE'
+  def admin_can_override?
+    user_type == :admin && ENV.fetch('OVERRIDE_OBSERVATION_STATUS', nil)&.upcase == 'TRUE'
```
tsubik, line 230: What is user type here, and where is it set to be `:admin`?

tsubik: Looks like for admin panel validation of status changes will never be invoked, right?

app/models/observation.rb
```diff
@@ -240,6 +247,10 @@ def set_centroid
 
   def update_operator_scores
     ScoreOperatorObservation.recalculate! operator if operator.present?
+    return unless operator_id_was == operator_id
+
+    old_operator = Operator.find_by id: operator_id_was
+    ScoreOperatorObservation.recalculate! old_operator if old_operator.present?
```
tsubik, line 253: Changing operator is still disabled, so how this code could be invoked? Or maybe you forgot to remove `disabled` from operator input. I thought operator should be changeable and that was the point of the feature?

app/admin/observation.rb
```diff
@@ -413,6 +413,7 @@ def self.observations
   end
 
   form do |f|
+    allow_override = ENV.fetch('OVERRIDE_OBSERVATION_STATUS', nil)&.upcase == 'TRUE'
```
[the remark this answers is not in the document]
santostiago: @tsubik yeah, this is supposed to be for them to temporarily edit observations that were created wrongly.
I'm guessing this will happen more times in the future.

app/admin/observation.rb
santostiago: @tsubik indeed... I'll check it. I think it was Rubocop that changed it.

app/models/observation.rb
santostiago: > Looks like for admin panel validation of status changes will never be invoked, right?

Yup.

app/models/observation.rb
santostiago: @tsubik mmm... I'm not sure what you're referring to. The operator can be modified (what cannot be is the type of observation, because I didn't want to check all the consequences and I don't think that should change).

spec/models/observation_spec.rb
```diff
@@ -71,28 +71,80 @@
 
 
   describe 'Validations' do
-    describe 'Status changes' do
+    describe '#status_changes' do
+      let(:country) { FactoryBot.create(:country) }
```
santostiago, line 75: I'm pretty sure I didn't write `FactoryBot`.. Could it be that Rubocop did it? (or maybe I'm going crazy :P )

app/models/observation.rb
tsubik: I meant, I have not seen setting user_type to admin anywhere that's why I'm asking if that's a mistake to check it here?

app/models/observation.rb
tsubik: I cannot modify operator, it has `disabled: operator` in the input. Select stays disabled in admin panel.

app/models/observation.rb
santostiago: @tsubik you're absolutely right.
In fact, the rules seemed to have been wrong before. We were just validating the transitions when it came from the observations tool.
I'm changing it.

santostiago:
@tsubik I modified it, but I'll wait for Sophie to know if this is what they want.
Previously, we were allowing admins to change the status of an observation to whatever they wanted. But I'm not sure if that should be the case.
Depending on what Sophie replies, I'll update the PR.

santostiago:
Ok, so we decided to just go with having `admins` and `bo_admins`

tsubik:
Alright, so many modifications are needed in that case :/

santostiago:
@tsubik ok, so:
- Admins can modify everything in the observations
- BO Managers can modify the status and admin comments
- Batch actions are hidden unless there's a flag

The only thing that was not very clear on Sophie's spec was if the admins could modify the status at will. I'm going to confirm that.

app/admin/observation.rb
```diff
@@ -35,28 +41,42 @@ def self.observations
...
+    resource.validation_status = Observation.validation_statuses['Needs revision']
+    notice = resource.save ?  'Observation approved' : 'Observation not modified'
+    redirect_to collection_path, notice: notice
   end
 
   member_action :start_qc, method: :put do
-    resource.update(validation_status: Observation.validation_statuses['QC in progress'])
-    redirect_to collection_path, notice: 'QC in progress for observations'
+    resource.validation_status = Observation.validation_statuses['QC in progress']
+    notice = resource.save ?  'Observation approved' : 'Observation not modified'
```
tsubik, line 78: Why all messages are changed to "observation approved"?

app/admin/observation.rb
```diff
@@ -82,60 +102,48 @@ def self.observations
...
   end
 
-  batch_action :move_to_qc_in_progress, confirm: 'Are you sure you want to start QC for all these observations?' do |ids|
-    batch_action_collection.find(ids).each do |observation|
-      observation.update(validation_status: Observation.validation_statuses['QC in progress']) if observation.validation_status == 'Ready for QC'
-    end
-    redirect_to collection_path, notice: 'QC started'
-  end
+  # Bulk actions should be available only if this env flag is set to `true`
+  if ENV.fetch('BULK_EDIT_OBSERVATIONS', 'false').upcase == 'TRUE'
```
tsubik, line 106: If we don't need that, I would vote to just remove it and not take space here ;)

app/admin/observation.rb
```diff
@@ -413,68 +420,123 @@ def self.observations
   end
 
   form do |f|
-    operator   = object.operator_id.present? ? true : false
-    fmu        = object.fmu_id.present? ? true : false
-    government = object.government_ids.present? ? true : false
+    allow_override = current_user.user_permission.user_role == 'admin'
```
tsubik, line 423: The form got very big and confusion, but maybe that's ok. Just thinking if for `bo manager` we can only show necessary fields if there are any and the fields that the user can change. Not sure if that is a good idea

app/admin/observation.rb
santostiago: Mmmm.. right. I'll change them to the previous one.

app/admin/observation.rb
santostiago: Sophie said it shouldn't be done. But I'd bet it's gonna be needed again in a year or so.
Do you think we should just get rid of it?

app/admin/observation.rb
santostiago: @tsubik I thought of it, but they need to see those in order to approve or not and to comment.

tsubik: Ok. Good point

app/admin/observation.rb
tsubik: Then, maybe better to keep them without disabling?

tsubik: But keeping code we should also maintain it, and we should not maintain something that might be needed in future, but probably don't, so either we remove it or keep it. That's my opinion.

santostiago: As discussed with Sophie, let's keep this hidden under an ENV var.

santostiago:
Going to rebase it now.

app/admin/observation.rb
```diff
@@ -82,60 +102,48 @@ def self.observations
...
   end
 
-  batch_action :move_to_qc_in_progress, confirm: 'Are you sure you want to start QC for all these observations?' do |ids|
-    batch_action_collection.find(ids).each do |observation|
-      observation.update(validation_status: Observation.validation_statuses['QC in progress']) if observation.validation_status == 'Ready for QC'
-    end
-    redirect_to collection_path, notice: 'QC started'
-  end
+  # Bulk actions should be available only if this env flag is set to `true`
+  if ENV.fetch('BULK_EDIT_OBSERVATIONS', 'TRUE').upcase == 'TRUE'
```
santostiago, line 106: @tsubik I ended up changing it to be on by default and we add the var with `false` in the server.
Doing it the other way was proving to be complicated to test (I was trying to remove the constant `Admin::ObservationsController` and then reloading the file, but I'd have to reload it using the AA loader and it wasn't working)

tsubik: ok

app/admin/observation.rb
tsubik: Those nested selects do not work properly, the list is not filtered

image

tsubik:
> * BO Managers can modify the status and admin comments

Hey @santostiago, but I see there are no valid status transitions for `bo_manager`.

app/admin/observation.rb
santostiago: @tsubik I didn't understand the screenshot. Which fields do not work? I tested it locally and it's filtering by the right field.

tsubik: Let me check it once again, as it was long time ago and I don't remember what was not working. Yeah, the screenshot could be better as it's not showing much.

tsubik: @santostiago both Operator and FMU selects do not work correctly 
image

santostiago:
@tsubik we're sending the requests from the observations admin page as admin, so it will have the same permissions (this can be improved, because we should instead check who made the request and set the permissions there, but I'll do it in another PR).
As for the filters, they seem to be working: I just filtered by one country, and the only operators shown are for that country. There's an error though which I'm going to check now: we're sending only the first page (20 results). I'll check why now.

app/admin/observation.rb
santostiago: @tsubik ohh.. now I get what you mean. The filtered query is not being used, correctly, no. Because we're using a translation table, it seems like ransack cannot interpret it properly.

For example, with the `Democratic Republic of Congo` selected, when I search for the operator `Andre` the ajax query made is:
`http://localhost:3000/admin/producers?order=operator_translations.name_desc&q[groupings][0][m]=or&q[groupings][0][operator_translations.name_contains]=andre&q[combinator]=and&q[country_id_eq]=7`

I'm not sure what these 2 params do:
- `q[groupings][0][m]=or`
- `&q[combinator]=and`
I need to check ransack. I'm also not sure if the right way to query the name is by using `q[groupings][0][operator_translations.name_contains]=andre`.  But this is done by the gem. It might be easier if I just manage to change the controller of the `Operator` AA resource to handle this properly.

tsubik: I think there is a task to remove name from translated fields for operator (maybe that would help here), not sure if FMU as well.

santostiago: it would solve it, yes. (I just dug into ransack code, and the issue is that when there's a query with groupings, ransack searches for the field name in his context - and he doesn't have the translations in his context).

I made a quick hack to fix it. Not an ideal solution, but I wouldn't waste more time on this.

santostiago:
@tsubik can you review? The solution is a bit hacky, but I didn't want to waste more time with this.

app/admin/operator.rb
```diff
@@ -280,6 +282,24 @@
   end
 
   controller do
+    def index
```
tsubik, line 285: Problem is not only for operator but for many other attributes that are using this component, category, subcategory and others. Let me think about it.

tsubik: With hotwire that would be easy to just refresh form on every country select change or other select changes.

tsubik (approved):

PR #364, opened 2022-11-17
Chore/cleanup jsonapi

tsubik:


app/resources/v1/country_resource.rb
```diff
@@ -13,7 +13,6 @@ class CountryResource < JSONAPI::Resource
     has_many :required_operator_documents
     has_many :required_gov_documents
     has_many :governments
-    has_many :monitors
```
santostiago, line 16: Aren't we using this in the apps? (isn't there any request on the observations tool or the FE that does countries?include=monitors)?

app/resources/v1/operator_document_country_history_resource.rb
```diff
@@ -3,7 +3,6 @@
 module V1
   class OperatorDocumentCountryHistoryResource < OperatorDocumentHistoryResource
     has_one :required_operator_document_country
-    has_many :documents
```
santostiago, line 6: Mmm... isn't this fetched in the FE?
To show the previous files uploaded?

app/resources/v1/operator_document_country_resource.rb
```diff
@@ -3,7 +3,6 @@
 module V1
   class OperatorDocumentCountryResource < OperatorDocumentResource
     has_one :required_operator_document_country
-    has_many :documents
```
santostiago, line 6: Don't we include this to display the documents? Do we make another request to the `documents` endpoint?

app/resources/v1/operator_document_fmu_history_resource.rb
```diff
@@ -3,7 +3,6 @@
 module V1
   class OperatorDocumentFmuHistoryResource < OperatorDocumentHistoryResource
     has_one :required_operator_document_fmu
-    has_many :documents
```
santostiago, line 6: And here

app/resources/v1/operator_document_fmu_resource.rb
```diff
@@ -3,7 +3,6 @@
 module V1
   class OperatorDocumentFmuResource < OperatorDocumentResource
     has_one :required_operator_document_fmu
-    has_many :documents
```
santostiago, line 6: Same here

config/routes.rb
```diff
@@ -28,43 +28,43 @@
 
     scope '(:locale)', locale: /en|fr/ do
       # Resources
-      jsonapi_resources :about_page_entries, only: [:index, :show] do; end
```
santostiago, line 31: Mmmm.. I don't think this is a good idea (even though the way it's done it's a hack).
But if the block isn't specifically defined as empty, the `jsonapi_resources` will generate a lot of sub-routes that we don't need.

spec/integration/v1/categories_spec.rb
```diff
@@ -4,24 +4,6 @@ module V1
   describe 'Categories', type: :request do
     it_behaves_like "jsonapi-resources", Category, {
       show: {},
-      create: {
```
santostiago, line 7: Mmmm.. why the removal of all these tests? Observations tool users can still perform CRUD on all these resources, right? Why remove the specs?

santostiago (commented):
I have a couple of questions.
But in general, it is very DRY and cleaner.

santostiago (commented):
I have a couple of questions.
But in general, it is very DRY and cleaner.

santostiago (commented):
I have a couple of questions.
But in general, it is very DRY and cleaner.

santostiago (commented):
I have a couple of questions.
But in general, it is very DRY and cleaner.

santostiago:
(ok, GH got crazy and repeated the same stuff a few times :D )

app/resources/v1/operator_document_country_resource.rb
tsubik: What are `documents` how jsonapi could know if there is no model. The whole problem with configuration was those weird `do; end` blocks in `routes.rb` when I removed that, the project was not even compiling showing the errors like `documents` do not exist, `monitor` model does not exists etc. That's why I removed all those mistakes, there is no such thing as `document` as there is no corresponding model.

app/resources/v1/country_resource.rb
tsubik: There is no model `monitor` that never worked and errors were silenced by `do; end` blocks in routes

config/routes.rb
tsubik: No idea where this hack is coming from, only I know it's silencing wrongly configured resources errors. I'm removing it and doing things as in documentation.

spec/integration/v1/categories_spec.rb
tsubik: I removed all routes that are not used. Categories are not created in obs tool.

tsubik: The code was commented out a long time ago.

tsubik: Also removed all not necessary API routes.

app/resources/v1/country_resource.rb
tsubik: I have no idea why those empty blocks in `routes.rb` silenced many resource configuration errors, but I always wanted to remove those blocks as it looked just weird.

config/routes.rb
tsubik: Do you remember why `do; end` were added to all routes?

tsubik: Ok, you are right, it's adding relationships routes and the only way to remove them is to add empty block. I will add them back as don't like that many endpoints. Thanks a lot :+1:

santostiago (approved):

PR #365, opened 2022-11-23
Fix redirect back in admin panel

tsubik:
I just discovered that in Rails 5.1 there is no `redirect :back` and specs were not catching any of those mistakes. I'm adding missing specs and fixing all places to have proper back redirection.

### DIFF
--- a/spec/controllers/admin/observation_documents_controller_spec.rb
+++ b/spec/controllers/admin/observation_documents_controller_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+# Base specs for admin default actions are done in active_admin_spec.rb
+RSpec.describe Admin::EvidencesController, type: :controller do
+  let(:admin) { create(:admin) }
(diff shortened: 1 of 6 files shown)


spec/controllers/admin/observation_documents_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
```
santostiago, line 1: @tsubik mmm.. this spec (and the ones after) is not really related to the PR, right?
It's just something that wasn't done before, right?

santostiago (approved):

spec/controllers/admin/observation_documents_controller_spec.rb
tsubik: Those are actually related. See that is testing the member actions `really_destroy` which were failing as were broken. The test is just adding those missing specs.

PR #369, opened 2022-11-24
Fix saving operator document annexes

tsubik:
Fixing saving of operator document annexes after rails upgrade. Adding specs that should be there before :P

### DIFF
--- a/spec/integration/v1/operator_document_annexes_spec.rb
+++ b/spec/integration/v1/operator_document_annexes_spec.rb
@@ -1,0 +1,54 @@
+require 'rails_helper'
+
+module V1
+  describe 'OperatorDocumentAnnex', type: :request do
+    let(:operator_document) { create(:operator_document_fmu) }
+    let(:document_data) {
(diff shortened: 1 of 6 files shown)


spec/integration/v1/operator_document_annexes_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  describe 'OperatorDocumentAnnex', type: :request do
+    let(:operator_document) { create(:operator_document_fmu) }
+    let(:document_data) {
+      "data:application/pdf;base64,#{Base64.encode64(File.read(File.join(Rails.root, 'spec', 'support', 'files', 'doc.pdf')))}"
+    }
+    let(:operator_document_annex) {
+      create(:operator_document_annex, user: operator_user, operator_document: operator_document)
+    }
+
+    before {
```
santostiago, line 13: This is the kind of stuff that I usually do with `transient" :P

santostiago (approved):
👍

spec/integration/v1/operator_document_annexes_spec.rb
tsubik: Hmm, I think this is just one time hack

PR #370, opened 2022-11-24
Fix docs position migration

tsubik:
Fixing migration


db/migrate/20221112185114_add_position_to_documents.rb
```diff
@@ -17,10 +17,10 @@ def change
     end
 
     RequiredGovDocument.group(:country_id).pluck(:country_id).each do |country_id|
-      RequiredOperatorDocumentGroup.find_each do |rodg|
+      RequiredGovDocumentGroup.find_each do |rgdg|
```
santostiago, line 20: 🤦

santostiago (approved):

PR #371, opened 2022-11-25
Chore/even more specs

tsubik:


app/controllers/v1/required_operator_document_groups_controller.rb
```diff
@@ -1,10 +0,0 @@
-# frozen_string_literal: true
```
santostiago, line 1: @tsubik we don't have requests for this one?

spec/controllers/admin/observations_controller_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+      before { put :start_qc, params: {id: observation.id} }
+
+      it 'is successful' do
+        expect(flash[:notice]).to match('QC in progress for observation')
+        expect(observation.reload.validation_status).to eq('QC in progress')
+      end
+    end
+  end
+
+  describe 'batch actions' do
```
santostiago, line 79: I'll remove these in the PR to implement `bo_managers` 😞 
Sophie said they don't need batch actions (well, maybe it could be shown depending on an env var)

santostiago (approved):
Much needed! 👍

app/controllers/v1/required_operator_document_groups_controller.rb
tsubik: No we don't. We only request documents and sideloading required documents with groups. I was thinking if to either add specs for those API endpoints I removed or just delete them, and decided to remove them as fewer endpoints less problems.

spec/controllers/admin/observations_controller_spec.rb
tsubik: Hmm, if they don't need batch actions then yeah better to remove, but all batch actions? I can remove those specs but just in case will keep the commit with them in the history.

tsubik: Also, are we talking about batch actions for observations only? Or they don't need them for other models as well?

tsubik:
Trying to bring up coverage to more than 90% before merging, Rails 52.

tsubik:
More than 90% now, so merging this.

PR #374, opened 2022-11-27
feat(Admin): Add links to download CSV in EN and FR

santostiago:
[Task](https://www.pivotaltracker.com/story/show/181285671)

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: 181285671

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/config/initializers/active_admin.rb
+++ b/config/initializers/active_admin.rb
@@ -147,7 +147,11 @@ ActiveAdmin.setup do |config|
   end
 
   # Set default locale for active admin
-  config.before_action :set_admin_locale
+  def set_admin_locale
+    return I18n.locale = params[:locale] if params[:locale] && params[:format] == 'csv'
+
+    super
+  end
 
   # == Localize Date/Time Format
   #
@@ -235,25 +239,10 @@ ActiveAdmin.setup do |config|
   config.view_factory.header = CustomAdminHeader
 
   # == Download Links
-  #
-  # You can disable download links on resource listing pages,
-  # or customize the formats shown per namespace/globally
-  #
-  # To disable/customize for the :admin namespace:
-  #
-  #   config.namespace :admin do |admin|
-  #
-  #     # Disable the links entirely
-  #     admin.download_links = false
-  #
-  #     # Only show XML & PDF options
-  #     admin.download_links = [:xml, :pdf]
-  #
-  #     # Enable/disable the links based on block
-  #     #   (for example, with cancan)
-  #     admin.download_links = proc { can?(:view_download_links) }
-  #
-  #   end
+
+  config.namespace :admin do |admin|
+    admin.download_links = [:csv, :json]
+  end
 
   # == Pagination
   #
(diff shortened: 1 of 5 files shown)


santostiago:
@tsubik I finished this one late and didn't check if all the specs are passing.
If they're failing, I'll fix them tomorrow, but you can check the PR nevertheless

app/controllers/concerns/translatable.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Translatable
```
tsubik, line 3: I'm not a fan of this as we're moving away from the default way of adding extra includes in Active Admin which is overriding `scoped_collection` method. If I want to add extra includes I would not know what translate method stands for I would probably just add `scoped_collection` as in AA docs.

app/controllers/concerns/translatable.rb
```diff
@@ -0,0 +1,25 @@
...
+  def translate(relationships = [], with_self = true)
+    controller do
+      before_action :set_locale, only: :index
+
+      def set_locale
```
tsubik, line 8: I think we could move setting this before_action to active admin initializers as this only works when you provide locale param explicit so we can make this work for all resources, and I would keep `scoped_collection` as there were before.

app/controllers/concerns/translatable.rb
santostiago: @tsubik what you're suggesting is not to have the `scoped_collection` there? 
It's a bunch of repeated code in most AA resources.

app/controllers/concerns/translatable.rb
santostiago: Mmmm.. let me check it

app/controllers/concerns/translatable.rb
tsubik: Yes, I think more code but easier to undestand could be beneficial, as when someone looks at `scoped_collection` then it's obvious what it does as this is default mechanism of AA. In comparison API `translate([[operator: :translations], [fmu: :translations]], false)` is very criptic.

tsubik: I'm just not sure if it's worth it to DRY it, as it is making this more complex and harder to customize scoped collection later, let's say we want to sideload different relations if it's CSV download vs HTML page as I did that many times.

config/initializers/active_admin.rb
```diff
@@ -307,3 +292,21 @@
   # Add Filters Extensions
   ActiveAdmin::BaseController.include ActiveAdmin::FilterSaver::Controller
 end
+
+# Modifies the display of the CSVs so that it shows on to download an English and a French versions.
+module ActiveAdmin::ViewHelpers::DownloadFormatLinksHelper
```
tsubik, line 297: Nice :+1:

app/controllers/concerns/translatable.rb
tsubik: Actually, it could be not worth it to try sideload only `I18n.locale` translations with custom `where` or some `joins`.

santostiago:
@tsubik I changed it no to have the `concern`.

tsubik:
Great. All look good except a couple of failing specs. I don't know why those are failing (checked them briefly). Please make sure all tests are passing and we can merge this.

santostiago:
The specs are failing because of the batch actions in observations. 
Merging the PR that changes the batch actions for observations should fix it. If not, I'll check tomorrow why it's happening.

tsubik (approved):
Alright, let's wait for other PR to be merged and see if all specs are passing then this is good to merge :+1:

santostiago:
Rebased

santostiago:
It still has failing specs. I'm going to check it.

PR #377, opened 2022-12-07
Feature/gov required docs parent

tsubik:
- Adding parent to required gov document groups, allowing only one level deep (in admin panel parent could be chosen only from top level)

Discussion link: https://www.pivotaltracker.com/story/show/181285821/comments/234612886


santostiago (approved):
All good.

PR #379, opened 2022-12-09
Change how expiration works for documents based on different statuses

tsubik:
Not required documents should "expire" to "not provided" state. This PR also contains a quick fix for documents that expired but should be not provided.


santostiago (approved):
👍

santostiago (approved):
It's still draft, but it looks fine.

PR #381, opened 2022-12-12
Fixing back Redirectable

tsubik:
Little hacky, but looks like it works.

Story: https://www.pivotaltracker.com/story/show/183760908

### DIFF
--- a/app/controllers/concerns/back_redirectable.rb
+++ b/app/controllers/concerns/back_redirectable.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module BackRedirectable
-  def back_redirect
-    controller do
+  def self.extended(base)
+    base.controller do
       def update
         update! do |success, failure|
           success.html do
--- a/app/controllers/concerns/back_redirectable.rb
+++ b/app/controllers/concerns/back_redirectable.rb
@@ -4,21 +4,11 @@ module BackRedirectable
   def self.extended(base)
     base.controller do
       def update
-        update! do |success, failure|
-          success.html do
-            redirect = params[:return_to] || collection_path
-            redirect_to redirect, notice: "#{resource.model_name.human} was successfully updated."
-          end
-        end
+        update! { params[:return_to] || collection_path }
       end
 
       def create
-        create! do |success, failure|
-          success.html do
-            redirect = params[:return_to] || collection_path
-            redirect_to redirect, notice: "#{resource.model_name.human} was successfully created."
-          end
-        end
+        create! { params[:return_to] || collection_path }
       end
     end
   end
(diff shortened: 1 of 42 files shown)


app/controllers/concerns/back_redirectable.rb
```diff
@@ -1,36 +1,28 @@
...
+  def self.extended(base)
+    base.controller do
       def update
-        update! do |success, failure|
-          success.html do
-            redirect = session.delete(:return_to) || collection_path
-            redirect_to redirect, notice: "#{resource.model_name.human} was successfully updated."
-          end
-        end
+        update! { params[:return_to] || collection_path }
```
santostiago, line 7: This won't return to the previous page on failure?

app/controllers/concerns/back_redirectable.rb
```diff
@@ -1,36 +1,28 @@
...
+        create! { params[:return_to] || collection_path }
       end
+    end
+  end
 
+  def form(options = {}, &block)
+    if block.present?
+      extended = Proc.new do |f|
+        return_to = request.params[:return_to] || request.referer
+        f.hidden_field :return_to, name: :return_to, value: return_to if return_to.present?
```
santostiago, line 20: shouldn't it be also disabled so that it's not sent in the form params?

tsubik:
It should be sent with forms params, that the trick to save it between
validation errored requests. As for redirect shortcut I took it from
inherited resources documentation so it should be good.

app/controllers/concerns/back_redirectable.rb
```diff
@@ -1,36 +1,28 @@
...
-        session[:return_to] ||= request.referer
-        super
-      end
-
-      def new
-        session[:return_to] ||= request.referer
-        super
-      end
-
+  def self.extended(base)
```
santostiago, line 4: this gets evaluated just by extending the module?

santostiago (approved):
Not sure I understood it well :P

tsubik:
Yeah, first time I use that to be honest.

tsubik:
Sorry looks like responding via email is not great as it responds to main thread.

app/controllers/concerns/back_redirectable.rb
tsubik: Yes.

app/controllers/concerns/back_redirectable.rb
tsubik: this must be sent back to controller (saved between requests) via params in this case. Otherwise, it won't work after validation errors. That's why I first check `params[:return_to]` if there is a value.

app/controllers/concerns/back_redirectable.rb
tsubik: This shortcut is described here https://github.com/activeadmin/inherited_resources#overwriting-actions

PR #386, opened 2022-12-21
Chore/refactor fmu properties

tsubik:
Removing unnecessary queries and quirks. Removing `properties` from Fmu as not sure why it was added if there are duplicated `geojson.properties`


santostiago (approved):
I didn't test it but it looks good.
But don't we use the properties in any other place?
Also, do we still need the geometry? Can't we just use the geojson?

tsubik:
Well, maybe some sort of trigger would be better, but if we dynamically
change geojson to geometry then fmu tiles request will be 2, 3 times slower
at least at my machine it worked like that.

PR #387, opened 2022-12-21
Fix old observations lat/lng

tsubik:
Task to fix old observations lat/lng. The task will find all observations for which point with swapped lat/lng is closer to FMU centroid than saved one.

### DIFF
--- a/lib/tasks/observations.rake
+++ b/lib/tasks/observations.rake
@@ -99,6 +99,7 @@ namespace :observations do
     puts "#{results.count} results with swapped lng lat inside fmu"
     puts "Observation ids: #{results.map { |r| r['id'] }.join(', ')}"
     ActiveRecord::Base.transaction do
+      Observation.skip_callback(:save, :after, :update_fmu_geojson)
       results.each do |res|
         observation = Observation.find(res['id'])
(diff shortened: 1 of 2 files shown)


lib/tasks/observations.rake
```diff
@@ -64,4 +64,53 @@ namespace :observations do
...
+          inner join fmus f on f.id = o.fmu_id
+          left join operator_translations ot on ot.operator_id = o.operator_id and ot.locale = 'en'
+          where f.geojson is not null
+        ) as temp
+      ) as temp2
+      where swapped_closer_to_centroid = true
+    SQL
+
+    results = ActiveRecord::Base.connection.execute(query).to_a
+    puts "#{results.count} results with swapped lng lat inside fmu"
```
santostiago, line 99: Wouldn't it be better to just return here if it's a dry run?

santostiago (approved):

lib/tasks/observations.rake
tsubik: For me it's better to run everything in transaction in this way I'm at least testing if the following code works. Sometimes I'm also adding additional assertions after running the script that's why using transaction by default.

PR #388, opened 2022-12-22
Add portuguese

tsubik:
+ new task to fill translations table with fallback entries.

### DIFF
--- a/config/initializers/locale.rb
+++ b/config/initializers/locale.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
-I18n.available_locales = [:en, :fr, 'zh-CN', :ja, :ko, :vi]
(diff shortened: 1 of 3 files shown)


config/initializers/locale.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
-I18n.available_locales = [:en, :fr, 'zh-CN', :ja, :ko, :vi]
+I18n.available_locales = [:en, :fr, :pt, 'zh-CN', :ja, :ko, :vi]
```
santostiago, line 3: it's `pt`? Not `pt-pt`  or `pt-br`?

tsubik:
Well, I didn't ask tbh, just did the same as for English or French. I think
that does not matter that much, right?

santostiago (approved):
👍

tsubik:
At least in our case.

PR #391, opened 2023-01-11
Backoffice: enhance displaying versions

tsubik:
Display error message instead 500 page when paper trail reify did not work. We cannot run `reify` on `create` paper trail versions (according to documentation and also that returns just `nil`), that's why we only try to reify `update`, `destroy` events.

When object let's say Producer is created then `Versionate` sidebar will show that there are no other versions for this object. After every update there will be a new version, version = 0 is the first version.


tsubik:
Moving this to draft for now. I need to refactor showing up first version.

tsubik:
Hmm, I dont' see "Draft" option anymore, not sure why.

santostiago (approved):
👍

PR #395, opened 2023-01-18
Fix notify observer status email lang and infraction text

tsubik:


santostiago (approved):
Shouldn't we do it for all the emails?

tsubik:
Most of the emails are only in English, we can use user's language only for this one right now. Sophie is currently cleaning up, changing texts of all emails, so there will be changes in the future.

PR #397, opened 2023-01-19
Fix email sending to multiple recipients

tsubik:
If we want to send email to multiple recipients, then always use blind carbon copy (BCC).


santostiago (approved):
Didn't test it, but the code looks fine 👍

PR #398, opened 2023-01-19
Feat/operator name uniqueness

tsubik:
When creating new operators, we want to make sure the name is unique.


tsubik:
Until all duplicates in the database are removed, uniqueness is only on create.

santostiago (approved):
👍

PR #401, opened 2023-01-22
fix(Admin Translations) Add translated fields

santostiago:
Replaces hardcoded string in the AA with I18n and adds the french translations.

## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: |
-          sudo apt update
+          sudo apt update --fix-missing
           sudo apt-get -yqq install libpq-dev build-essential libcurl4-openssl-dev gdal-bin
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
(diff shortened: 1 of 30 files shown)


santostiago:
@tsubik there are still a bunch of models to translate, as well as enums and places with hardcoded strings (still didn't check how it work with globalized attributes).
I'll try to finish it during the week.

santostiago:
@tsubik there are still a few places missing: some JS, some partials and some weird spots that I may have missed.
But let's merge it and ask feedback from Sophie. There shouldn't be much more left, and I'll address it in another PR (also, I'm going crazy doing this :P)

santostiago:
Oh.. and there's also the translations of the enums. Any suggestion on how to do that?

tsubik (approved):

tsubik:
For enums if we need to add translations we could add simple PORO models for each enum (example here https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/models/concerns/enum_model.rb, https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/models/open_call_status.rb) to encapsulate some logic. In locale files I already see one enum translated https://github.com/wri/fti_api/blob/develop/config/locales/en.yml#L37

PR #406, opened 2023-01-26
Backoffice: fix observation links in gov entity page

tsubik:
Fix observation links https://www.pivotaltracker.com/story/show/184318922


santostiago (approved):
Ups...

PR #408, opened 2023-02-01
Add simple middleware for health check

tsubik:
This will include more checks in the future.


santostiago (approved):
Mmmm.. isn't there a gem for this?
(And why the commented Redis check?)

tsubik:
As we don't use redis yet. Will uncomment it when we use action jobs with sidekiq, but I'm still not sure if sidekiq for this app is not overkill.

tsubik:
Also, no need to add gem for a couple lines of code.

PR #412, opened 2023-02-03
Feature/api return proper forest types

tsubik:
Use forest_type enum (not extra geojson_label) in API and also add a new property to geojson which will be used by frontend to correctly color layers.

After merging this we need to update geojson properties

```
bundle exec rails update:geojson
```

After frontend changes are deployed, we can remove geojson_label completely.

Tested with frontend already.


santostiago (approved):
👍

PR #416, opened 2023-02-08
Fix admin translations

tsubik:
Adding simple specs to catch some of the missing translations in back-office and fixing a couple of them.


santostiago (approved):
👍

PR #420, opened 2023-02-16
Chore/jsonapi upgrade to 0.9.12

tsubik:
Upgrading jsonapi resources to newest 0.9 version which has Rails 6.1 support.

Note: I tried both master and 0.10 branch of jsonapi resources and always had different issues. I'm tired of this library, but creating API from scratch using for example jsonapi serializer is a big task at this point.

I tested that with both frontend apps using automated Cypress tests and all looked good. Will push this to staging soon to have it there for a while.

### DIFF
--- a/app/resources/v1/base_resource.rb
+++ b/app/resources/v1/base_resource.rb
@@ -1,9 +1,5 @@
 module V1
   class BaseResource < JSONAPI::Resource
     abstract
-
-    def custom_links(_)
-      { self: nil }
-    end
   end
 end
--- a/config/initializers/jsonapi_resources.rb
+++ b/config/initializers/jsonapi_resources.rb
@@ -8,6 +8,7 @@ JSONAPI.configure do |config|
   config.resource_cache = Rails.cache
   config.always_include_to_one_linkage_data = false
   config.warn_on_missing_routes = false
+  config.default_exclude_links = :default
 
   # Metadata
   # Output record count in top level meta for find operation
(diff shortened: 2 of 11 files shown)


app/resources/v1/base_resource.rb
```diff
@@ -1,9 +1,5 @@
 module V1
   class BaseResource < JSONAPI::Resource
     abstract
-
```
santostiago, line 4: We don't need this anymore? It doesn't include the custom links by default?

config/initializers/jsonapi_resources.rb
```diff
@@ -7,6 +7,8 @@
   config.maximum_page_size = 3000
   config.resource_cache = Rails.cache
   config.always_include_to_one_linkage_data = false
+  config.warn_on_missing_routes = false
+  config.default_exclude_links = :default
```
santostiago, line 11: Nice!

santostiago (approved):

santostiago:
What issues did you run into when trying to move to 0.10?

config/initializers/jsonapi_resources.rb
tsubik: Yes, here is the `default_exclude_links` that removes links. Although, as this gem is so unstable across different versions it's always better to check if it works ;]

tsubik:
I don't even want to start talking about it. 0.10 is rewrite of crucial components and in that version or maybe master, don't even remember now, either STI works differently, or stops including nested relations or any relations in `included` (on master it started including relationships to main objects causing most of frontend to have circular reference errors). I would love to rewrite API in something different and forget about that gem lol or we can wait and see if there are any signs of improvements for that gem (as right now I can say it's almost unmaintained)

PR #425, opened 2023-03-10
Fix/stats generation forest types

tsubik:
Fixing scheduled operator document statistic generation and small refactoring for forest types.


santostiago (approved):
👍

PR #431, opened 2023-04-02
fix(Translations) Correct translations

santostiago:
Added dynamic translations to the backoffice. Corrected others that identified as faulty.

## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


app/admin/about_page_entries.rb
```diff
@@ -10,9 +10,9 @@
   permit_params :position, :code, translations_attributes: [:id, :locale, :title, :body, :_destroy]
 
   filter :position, as: :select
-  filter :translations_title_contains, as: :select, label: 'Title',
+  filter :translations_title_contains, as: :select, label: I18n.t('activerecord.attributes.about_page_entry.title'),
```
tsubik, line 13: here as well, label should be in proc or lambda. You changed that to proc in couple places but forgot in some. There are a only couple of them when I search phrase like "label: I18n"

tsubik: most are false positives, so only a handful of changes needed.

app/assets/javascripts/pagination.js
```diff
@@ -1,16 +1,17 @@
...
 
   var goto_page_val = $("#hidden_active_admin_per_page").val();
+  const jumpToPageText = $("#jump-to-page").data('value');
 
   $("#hidden_active_admin_per_page").remove();
 
   $("form.filter_form").prepend(' \
     <div class="input optional filter_form_field " id="goto_page_input"> \
-    <label for="page" class="label">Jump to page</label> \
+    <label for="page" class="label">' + jumpToPageText + '</label> \
```
tsubik, line 10: ahh I see, that string is not from active admin, it's a custom part. Well, I guess using some hidden field is the easiest (and best way :+1: ) I can also think about now to resolve that issue. If we need more texts in JS we can extend this to provide hash as hidden field data attribute to have more strings there.

tsubik (approved):
Great :100:

PR #432, opened 2023-04-04
Chore/rubocop

tsubik:
Upgrading Rubocop and moving to standard.rb (https://github.com/testdouble/standard).

Disabling Rails cops for now. I will upgrade that part in separate PR.


santostiago (approved):
🙈

PR #434, opened 2023-04-06
Chore/ruby upgrade

tsubik:
Staging is already running on 3.2.2 :tada:


tsubik:
Yeah, probably, never tried that in Ruby.

PR #436, opened 2023-04-12
Chore/rails 7

tsubik:


config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,4 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :marshal
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
santostiago, line 5: why?

santostiago (approved):
The changes are pretty small. Cool!

config/initializers/cookies_serializer.rb
tsubik: Because `marshal` does not work anymore, at least I have errors when saving filters in active admin. We should move to `json` after some time, `hybrid` is just temporal.

tsubik: Not mentioning brakeman ignored warnings when using marshal that could be silenced

PR #441, opened 2023-05-22
fix observations year filter

tsubik:
Observation report year should be used for the filter.


app/resources/v1/observation_resource.rb
```diff
@@ -59,13 +59,8 @@ class ObservationResource < BaseResource
...
 
     filter :years, apply: ->(records, value, _options) {
-      records.where("extract(year from observations.publication_date) in (#{value.map { |x|
-                                                                              begin
-                                                                                x.to_i
-                                                                              rescue
-                                                                                nil
-                                                                              end
-                                                                            }.join(", ")})")
+      years = value.map(&:to_i).compact.join(", ")
```
santostiago, line 62: If we get a string for the value of years, we'll search for year 0. Is that ok?

santostiago (approved):

app/resources/v1/observation_resource.rb
tsubik: I know, and I think it's fine.

tsubik: Actually, I will just reject zeros

PR #442, opened 2023-05-23
fix translation caching issues

tsubik:
Couple models were badly configured without `touch: true` for translations that was causing caching problems.


santostiago (approved):
👍

PR #443, opened 2023-05-23
Fix subcategories translations edit

tsubik:
Not sure why default scope clashes with `with_translations` but I don't have much time now to investigate this further. Ideally would be to get rid of all default scopes.

Now, all languages are visible.
image

### DIFF
--- a/app/admin/subcategory.rb
+++ b/app/admin/subcategory.rb
@@ -11,7 +11,7 @@ ActiveAdmin.register Subcategory do
 
   controller do
     def scoped_collection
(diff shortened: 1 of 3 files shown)


santostiago (approved):
Weird. It's just for this resource?

tsubik:
Looks like `includes(:translations).with_translations(I18n.locale)` is a bad combination, preventing getting all `object.translations`, only current locale one is returned. If we do use each separately, we get all translations. I think it would be better to actually remove default scope here.

PR #444, opened 2023-05-30
Couple changes for observation filter tree

tsubik:
Enhancing observation filters. https://www.pivotaltracker.com/story/show/183458561

### DIFF
--- a/app/controllers/v1/observation_filters_controller.rb
+++ b/app/controllers/v1/observation_filters_controller.rb
@@ -7,22 +7,6 @@ module V1
     skip_before_action :authenticate
 
     def tree
-      years = Observation.pluck(:publication_date).map { |x| x.year }.uniq.sort
-        .map { |x| {id: x, name: x} }
-
-      severities = [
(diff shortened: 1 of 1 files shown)


app/controllers/v1/observation_filters_controller.rb
```diff
@@ -44,15 +28,35 @@ def tree
 
     private
 
+    def years
+      Observation.published.pluck(:publication_date).map { |x| x.year }.uniq.sort.reverse
```
santostiago, line 32: I know this is the way it was done, but maybe an optimization would be to do this in SQL (it would be slightly faster)

santostiago (approved):
Looks good!

app/controllers/v1/observation_filters_controller.rb
tsubik: ohh, thanks for pointing this out. I will change it to SQL version, but more importantly `years` should not be taken from observation's publication_date but from observation report's publication_date. I changed the filter in observation_resource.rb the other day, but did not change the data source here.

PR #445, opened 2023-05-31
Add slug to operators

tsubik:
Add slug to operators, will be used by frontend app for better SEO.

https://www.pivotaltracker.com/story/show/184614863


santostiago (approved):
It looks good. But don't we need to update the routes?

tsubik:
No, I just added a filter, not using slug as id in the route, as I had problems with using hybrid approach (both id, and slug) as id in the route. The solution to override `find_by_key` does not work here (at least with this version of jsonapi-resources)

PR #450, opened 2023-07-03
Fix sending newsletter

tsubik:
The newsletter code will be eventually refactored, so I focused on just fixing it. We cannot use `bcc` without setting `to` when using SendGrid. I also read that BCC might not be great in terms of deliverability, that's why I decided to just send emails individually. We could do that with one API call with SendGrid API using `personalizations` object, but I preferred to use a method acceptable by any ActionMailer client.


tsubik:
Couple additions here. Allowing to resend only to some operators or users by using ENV variable, and send only to operators who signed publication authorization/approved status.

santostiago (approved):

PR #454, opened 2023-08-08
Feature/observation publication date cleanup

tsubik:
Observation publication date is only information, should be only updated if observation is published in any case, not only through API. Cannot be updated in the form.


santostiago (approved):
👍

PR #456, opened 2023-08-24
Feature/cms pages

tsubik:
Ability to add simple page with html content. We should migrate the rest of pages to use this simpler API.


app/admin/about_page_entries.rb
```diff
@@ -51,7 +51,7 @@ def scoped_collection
       f.input :position, hint: "leaving empty will assign last position"
       f.input :code, hint: 'must be "partners" for Partners and "donors" for Donors section'
     end
-    f.translated_inputs switch_locale: false do |t|
+    f.translated_inputs "Translations", switch_locale: false do |t|
```
santostiago, line 54: Shouldn't we localize this?

santostiago (approved):
All good!
I just left a comment regarding the localization of the word "Translations".

santostiago:
Oh.. and there's some issue with the Linter.

app/admin/about_page_entries.rb
tsubik: The thing is that it's not showing up anywhere. This gem is unmaintained and the param must be there but is unused.

tsubik: I removed it before, but realized that those options that are second parameter stopped working. Checked the gem code, and discovered that first parameter is not used.

PR #458, opened 2023-09-16
fix(Admin): Change filter by name on producers page

santostiago:
The filter by name on the producers page now uses the id of the producer selected.

## Pull request checklist

- [ ] Check the commit's or even all commits' message styles matches our requested structure.

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The search was done by name (which failed when the name had special characters).

Issue Number: #183458551

## What is the new behavior?
It now uses the selected producer's id.

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/admin/operator.rb
+++ b/app/admin/operator.rb
@@ -134,7 +134,7 @@ ActiveAdmin.register Operator, as: "Producer" do
   dependent_filters do
     {
       country_id: {
-        name_eq: Operator.pluck(:country_id, :name)
+        id: Operator.pluck(:country_id, :id)
       }
     }
   end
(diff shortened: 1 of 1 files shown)


tsubik (changes requested):
There are two issues with this, one minor one I think we can dismiss it that in `Current filters` section in Search status it says "Name equals [id]" I don't know if it's possible to change it to "ID equals [id]" with keeping the label `Name` for the filter.

Another issue is that dependent filters stopped working. Meaning if you change country then it should only show producers from that country, now there is JS error.

You should change `dependent_filters` section to be like

```
country_id: {
  id: Operator.pluck(:country_id, :id)
}
```

tsubik (approved):

PR #460, opened 2023-10-17
Use forked version of active_admin_paranoia

tsubik:
This fixes https://www.pivotaltracker.com/story/show/186270985

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ end
 gem "bootsnap", require: false
 
 # Active admin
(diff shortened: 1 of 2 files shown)


santostiago (approved):
What did you change on this version?

tsubik:
Here are the changes https://github.com/raihan2006i/active_admin_paranoia/compare/master...tsubik:active_admin_paranoia:master. Restore button was always visible and it should be only for destroyed!

PR #461, opened 2023-10-17
Fix observer endpoint

tsubik:
Re-add update action to observer endpoint.


santostiago (approved):
What was the issue with this one?

tsubik:
So I removed it by accident and adding this back for obs tool to be able to update observer details.

PR #465, opened 2023-10-30
NGO users can manage operators/gov entities

tsubik:
As in the title. Pivotal story here: https://www.pivotaltracker.com/story/show/185881210


santostiago (approved):
This is fine, but it's not enough.
You'll need to run a script to update the roles of the existing users.

tsubik:
Thanks. Yes, I know that. I had to change that script in other PR to not update `bo_manager` users role. I think keeping user_permissions in this way in DB is bad idea and would gladly remove that model in the future. Keep abilities static, and only make certain abilities updatable for `bo_managers` but not ALL user permissions for all users, that makes no sense for me to always load this model.

santostiago:
@tsubik we shouldn't do that.
The individual users' permissions can still be edited in the Back Office on the permission matrix.
Most of the times the general settings are fine, but there were requirements to change permissions for specific users.

santostiago:
Oh, you mean to keep this only for BO Managers.
Yeah, that fine. But where do you want to save them?

tsubik:
I thought that this access control should be directly in user's details in the admin panel not separately. What's more, I don't think we need those extra permissions to be in a separate table, the new data which won't have all permissions just couple overrides is not that heavy.

PR #466, opened 2023-11-06
Fix missing operator document validation

tsubik:
Validation for a `reason` why document is not applicable or `document_file` in other cases was missing.


spec/factories/operator_documents.rb
```diff
@@ -31,6 +31,7 @@
     start_date { Date.yesterday }
     type { "OperatorDocumentCountry" } # This can be overwritten by the children.
     document_file { build :document_file }
+    # status { "doc_not_provided" }
```
santostiago, line 34: ??

santostiago (approved):

spec/factories/operator_documents.rb
tsubik: Thanks, I forgot to remove that. Looks like status is set on create event based on if document_file is provided or the reason.

PR #470, opened 2023-11-16
Feature/protected areas

tsubik:
Manage protected areas in the back-office, to be able to update geometries ([PT story here](https://www.pivotaltracker.com/story/show/186300570))

Populate DB with initial import from GFW API.

```
bin/rails import:protected_areas
```

image


santostiago (approved):
I'm approving this one without having tested it, just to get it moving
It is very interesting, so later I'll check it better.
Keep in mind one thing that happened to me before: the validations in rails (using some geo gem) to the shapes can take a lot of time. In the previous form, I ended up removing this and having them only in the DB.

PR #473, opened 2023-11-30
More cleanup

tsubik:
- Remove comments (it is not used, and not planned to use)
- improve API
- updating gems, especially carrierwave to v3


santostiago (approved):
This `fetch_all` isn't used by the JsonAPIResources?

PR #474, opened 2023-12-07
Chore/db tasks

tsubik:
So far I was using external (not under source control) bash scripts to manage local database, but I was always thinking about improving the process. Here it is, introducing a couple new rake tasks to download database backups from remote servers and to recreate it to local copy.


santostiago (approved):
Really cool!

PR #477, opened 2023-12-21
Assign observers from observation report

tsubik:
Observation report should be the main source of observers for the observation. 

Main changes: 
- Changing observers for observation report should update all report's observations relations. 
- When setting a different observation report for existing observation, we need to make sure observers will be taken from new report.
- Disable changing observers for observation in admin panel
- Enable changing observers for observation report in admin panel 

We can keep the existing observation-observers relation for now, as changing this (removing that connection) could take a way more time. Changing API, and both applications to not seek observers in observation model but in observation report.

### DIFF
--- a/app/models/observation_report.rb
+++ b/app/models/observation_report.rb
@@ -25,7 +25,7 @@ class ObservationReport < ApplicationRecord
   # API creating report fails as it's not providing user. Thing to investigate
   belongs_to :user, inverse_of: :observation_reports, optional: true
   has_many :observation_report_observers, dependent: :destroy
-  has_many :observers, through: :observation_report_observers
+  has_many :observers, through: :observation_report_observers, after_add: :self_touch, after_remove: :self_touch
   has_many :observations, dependent: :destroy
 
   skip_callback :commit, :after, :remove_attachment!
@@ -33,9 +33,19 @@ class ObservationReport < ApplicationRecord
   after_restore :move_attachment_to_public_directory
   after_real_destroy :remove_attachment!
 
+  after_commit :sync_observation_observers
+
   scope :bigger_date, ->(date) { where("observation_reports.created_at <= ?", date + 1.day) }
 
-  def update_observers
-    self.observer_ids = observations.joins(:observers).distinct.pluck("observers.id") if observations.any?
+  private
+
+  def self_touch(_)
+    touch unless destroyed? || new_record?
+  end
+
+  def sync_observation_observers
+    observations.with_deleted.find_each do |observation|
+      observation.observers = observers
+    end
   end
 end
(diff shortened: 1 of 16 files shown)


santostiago:
Mmm... interesting...
But I thought there was the possibility of an observer to make an observation without a report. That is no longer valid?

app/models/observation_report.rb
```diff
@@ -25,17 +25,27 @@ class ObservationReport < ApplicationRecord
   # API creating report fails as it's not providing user. Thing to investigate
   belongs_to :user, inverse_of: :observation_reports, optional: true
   has_many :observation_report_observers, dependent: :destroy
-  has_many :observers, through: :observation_report_observers
+  has_many :observers, through: :observation_report_observers, after_add: :self_touch, after_remove: :self_touch
```
santostiago, line 28: mmmm.. I didn't know `after_add` and `after_remove` existed.
Cool!

app/models/observation_report.rb
```diff
@@ -25,17 +25,27 @@ class ObservationReport < ApplicationRecord
...
 
-  def update_observers
-    self.observer_ids = observations.joins(:observers).distinct.pluck("observers.id") if observations.any?
+  private
+
+  def self_touch(_)
+    touch unless destroyed? || new_record?
+  end
+
+  def sync_observation_observers
```
santostiago, line 46: I'm guessing in the future this won't be needed.

spec/factories/observations.rb
```diff
@@ -86,10 +47,10 @@
     severity { build(:severity, subcategory: subcategory) }
     operator { create(:operator, country: country) }
     observation_type { "operator" }
-    observers { build_list(:observer, 1) }
-    species { build_list(:species, 1, name: "Species #{Faker::Lorem.sentence}") }
+    # species { build_list(:species, 1, name: "Species #{Faker::Lorem.sentence}") }
```
santostiago, line 50: mmm.. I think we can remove this.

spec/factories/observations.rb
```diff
@@ -110,6 +71,23 @@
     end
   end
 
+  factory :gov_observation, class: "Observation" do
+    severity
+    country
+    governments { build_list(:government, 2) }
+    # species { build_list(:species, 1, name: "Species #{Faker::Lorem.sentence}") }
```
santostiago, line 78: same

santostiago (approved):
👍

tsubik:
Observation without report is only possible in "Created" state, which I would say should be named as "Draft" lol. In fact, we are missing all those required fields validations in the backend.

spec/factories/observations.rb
tsubik: Right. I forgot.

spec/factories/observations.rb
tsubik: :+1:

app/models/observation_report.rb
tsubik: Yes.

PR #478, opened 2024-01-03
Chore/activeadmin upgrade

tsubik:
Upgrading ActiveAdmin to v3 as well as upgrading all gems.


app/models/api_key.rb
```diff
@@ -21,4 +21,8 @@ class APIKey < ApplicationRecord
   def expired?
     DateTime.now >= expires_at || user.deactivated? || deactivated?
   end
+
+  def self.ransackable_attributes(auth_object = nil)
```
santostiago, line 25: mmm... what is this for?

app/models/fmu.rb
```diff
@@ -67,7 +67,7 @@ class Fmu < ApplicationRecord
   scope :filter_by_operators, ->(operator_ids) { joins(:fmu_operators).where(fmu_operators: {current: true, operator_id: operator_ids.split(",")}) }
   # this could also be done like: "id not in ( select fmu_id from fmu_operators where \"current\" = true)"
   # but it might break the method chaining
-  scope :filter_by_free, -> { where.not(id: FmuOperator.where(current: true).pluck(:fmu_id)).group(:id) }
+  scope :filter_by_free, -> { where.not(id: FmuOperator.where(current: true).select(:fmu_id)).group(:id) }
```
santostiago, line 70: 👍

config/initializers/active_record.rb
```diff
@@ -8,4 +8,16 @@ def structurally_incompatible_values_for_or(other)
         (Relation::CLAUSE_METHODS - [:having, :where]).reject { |m| send("#{m}_clause") == other.send("#{m}_clause") }
     end
   end
+
+  # After upgrading to ransack 4.0 we need to explicitly whitelist attributes
+  # In this project I prefer to only blacklist the attributes that are not allowed for couple models
+  class Base
```
santostiago, line 14: oh...

db/migrate/20210607152721_add_current_unique_index_to_score_operator_documents.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable all
```
santostiago, line 1: We could just add the folder migrations to the Rubocop exceptions.

lib/file_data_import/belongs_to_association.rb
```diff
@@ -62,7 +62,7 @@ def translations_attributes
 
     def extract_attributes(attrs)
       attrs.each_with_object({}) do |attribute, attributes|
-        value = raw_attributes["#{singular_name}__#{attribute}".to_sym]
+        value = raw_attributes[:"#{singular_name}__#{attribute}"]
```
santostiago, line 65: this also changed?

santostiago (approved):
👍

app/models/api_key.rb
tsubik: To blacklist all attributes of this model to not be usable in ransack queries.

db/migrate/20210607152721_add_current_unique_index_to_score_operator_documents.rb
tsubik: Yeah, that could be another option, but I thought we might need rubocop for new migrations. It has a lot of very useful warnings and hints to write better migrations.

lib/file_data_import/belongs_to_association.rb
tsubik: One of rubocop auto corrections.

PR #479, opened 2024-01-08
Chore/new seeds

tsubik:
Old seeds did not work. Instead of fixing them, I decided to use a subset of anonymized production data in the form of fixtures. Why fixtures? Because loading up fixtures in Rails is very quick and well-supported. 

This will also allow using better test data for e2e testing in both Portal and Observation Tool applications. For this, I also added custom e2e environment and new rake tasks (e2e.rake). Recreating database between tests is much quicker now.

To recreate fixtures, you have to load production database locally and run:

```
rails seeds:generate_fixtures
```


app/models/user.rb
```diff
@@ -38,7 +38,7 @@ class User < ApplicationRecord
   # Include default devise modules.
   TEMP_EMAIL_REGEX = /\Achange@tmp/
   PERMISSIONS = %w[operator ngo ngo_manager government]
-  PROTECTED_NICKNAMES = %w[admin superuser about root fti otp faq contact user operator ngo].freeze
+
```
santostiago, line 41: ehehe.. this thing was already there when I grabbed the project (like, a month after it started) and I always wanted to removed it and never did.
Now it's done :D

config/initializers/active_record_fixtures.rb
```diff
@@ -0,0 +1,85 @@
+require "active_record/fixtures"
```
santostiago, line 1: I have no idea what this file does :D

lib/tasks/e2e.rake
```diff
@@ -0,0 +1,64 @@
...
+        terminate_connections_to db_config["database"]
+        sh "dropdb --if-exists #{connection_config} #{db_config["database"]}"
+        sh "createdb #{connection_config} --template=#{template_db_name} #{db_config["database"]}"
+      end
+    end
+  end
+
+  private
+
+  def drop_e2e_db_template
```
santostiago, line 30: mmm.. cool.
How do the PG templates work?

lib/tasks/seeds.rake
```diff
@@ -0,0 +1,143 @@
...
+  OPERATORS = %w[ifo-interholco cfc sifco lorema siencam cib cft mokabi-sa afriwood-industries]
+  EXTRA_FMUS = %w[08-003 08-005 08-009]
+
+  DATE_FIELDS = %w[created_at updated_at deleted_at operator_document_updated_at operator_document_created_at publication_date response_date]
+
+  def initialize
+    namespace :seeds do
+      desc "Generate fixtures to use as development or e2e database. First run db:restore_from_file or db:restore_from_server to get latest data from production."
+      task generate_fixtures: :environment do
+        generate_for_model "Country"
```
santostiago, line 15: Maybe save an array of models and iterate over them calling `generate_for_model`?

santostiago: mmm... ok, I just saw later in the code that it won't work for all.

santostiago (approved):
Looks great.
When do you want to load the fixtures? To set up the dev environment?

config/initializers/active_record_fixtures.rb
tsubik: Sorry, I will add additional comment why it's here. When creating and then loading fixtures, I had foreign key exceptions, but in Rails 7 the exception has no details. In rails 7.1 this is fixed and you know which object is causing those errors, the error message is a way better. That's why I took that fragment of code from Rails 7.1 in form of extension here.

lib/tasks/seeds.rake
tsubik: Yeah, I started as you said, with array of models, then that changed to hash, but ended with this as this is the most flexible way.

PR #480, opened 2024-01-11
Feat/observation multiple evidences

tsubik:
Connect observation document with multiple observations, not only one.


db/migrate/20231222154005_create_observation_document_observations.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.belongs_to :observation_document, foreign_key: {on_delete: :cascade}, index: {name: "observation_documents_observations_doc_index"}, null: false
+      t.belongs_to :observation, foreign_key: {on_delete: :cascade}, index: {name: "observation_documents_observations_obs_index"}, null: false
+      t.index [:observation_document_id, :observation_id], unique: true, name: "observation_documents_observations_double_index"
+      t.timestamps
+    end
+
+    query = <<-SQL
+      INSERT INTO observation_documents_observations (observation_document_id, observation_id, created_at, updated_at)
+      SELECT
+        od.id, o.id, NOW(), NOW()
```
santostiago, line 13: mmm. wouldn't it be better to keep the `created_at` and `updated_at`?

santostiago (commented):
👍

db/migrate/20231222154005_create_observation_document_observations.rb
tsubik: Hmm, maybe you are right.

santostiago (approved):

PR #483, opened 2024-03-28
 Feature/admin operator documents perform qc

tsubik:
perform_qc


app/models/operator_document.rb
```diff
@@ -48,6 +49,7 @@ class OperatorDocument < ApplicationRecord
 
   validates :start_date, presence: {if: :document_file_id?}
   validates :expire_date, presence: {if: :document_file_id?}
+  validates :admin_comment, presence: {if: :doc_invalid?}
```
santostiago, line 52: maybe we'll have issues if someone wants to update an old doc that is invalid.
But ok, maybe they shouldn't update it.

santostiago (approved):
Looks great 👍

app/models/operator_document.rb
tsubik: True, I will double-check this if this is something to worry about.

PR #485, opened 2024-04-16
Feature/operator documents new notifications

tsubik:
Adding new notifications when operator document changes status.

PT: https://www.pivotaltracker.com/story/show/187126380


app/models/operator_document.rb
```diff
@@ -217,4 +219,18 @@ def reason_or_file
       errors.add(:base, "File must be present or reason when document is non applicable")
     end
   end
+
+  def notify_about_changes
```
santostiago, line 223: mmm.. from the task in Pivotal it looks like the goal is to notify when the documents are ready to be reviewed.
Like, the operators submitted them and the admins should review them.

But here the task tells the operators that their documents were reviewed.
Maybe the description of the task in Pivotal is not very clear.

santostiago (approved):

app/mailers/user_mailer.rb
```diff
@@ -31,7 +31,7 @@ def generate_reset_url(user)
 
   def generate_reset_token(user)
     token, hashed = Devise.token_generator.generate(User, :reset_password_token)
-    user.update(reset_password_token: hashed, reset_password_sent_at: DateTime.now)
+    user.update!(reset_password_token: hashed, reset_password_sent_at: DateTime.now) unless user.new_record? # workaround for mailer previews
```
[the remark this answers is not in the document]
tsubik: I think I had some validations errors here as in mailer's previews I'm not using persisted data, just `User.new` entity. I want to just get the reset token here without saving in db.

app/models/operator_document.rb
tsubik: I think you looked at wrong PT task. This one is "Set up email notifications to alert private sector users when their documents have been published or if they have been rejected", there is another one for notifying admins that observation is ready for review (I will add a new PR today for it).

PR #486, opened 2024-04-19
Fix missing observer in some observation mails

tsubik:
Sometimes, for example when backoffice admin updates observation through observation tool the observer is missing from `modified_user` as that user is of admin type and has no observer attached. In that case, we will use observer from the user who created observation if somehow that one is null then just use first observer from observation. 

PT: https://www.pivotaltracker.com/story/show/187390044


app/mailers/observation_mailer.rb
```diff
@@ -31,20 +31,20 @@ def observation_published(observation, user)
...
-    @observer = observation.modified_user.observer
+    @observer = observer
     mail to: user.email, subject: I18n.t(subject_i18n_key, observer: @observer.name, id: observation.id)
   end
 
   private
 
   def observer
-    @observation.modified_user.observer
+    @observation.modified_user.observer || @observation.user&.observer || @observation.observers.first
```
santostiago, line 47: and `modified_user` is never nil?

tsubik: Right, probably not, but just to be safe I will change it.

PR #488, opened 2024-04-23
Feature/remove monitor names translation

tsubik:
Monitor/Observer names should not be translated. I'm also removing `organization` attribute as looks like it is very obsolete.

PT: https://www.pivotaltracker.com/story/show/187407020


db/migrate/20240423144714_remove_observer_name_translation.rb
```diff
@@ -0,0 +1,28 @@
...
+  def up
+    # globalize migrate_data for drop_translation_table! does not work
+    add_column :observers, :name, :string
+    query = <<~SQL
+      UPDATE observers o
+        SET name = t.name
+      FROM
+        observer_translations t
+      WHERE
+        o.id = t.observer_id and t.locale = 'en'
```
santostiago, line 15: Maybe it's a good idea to run a query first just to confirm they all have names in English (and to check if the names are the same in French).

db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_11_30_113839) do
+ActiveRecord::Schema[7.1].define(version: 2024_04_23_144714) do
+  create_schema "tiger"
```
santostiago, line 14: mmm.. what is this for?

santostiago (approved):
👍

db/migrate/20240423144714_remove_observer_name_translation.rb
tsubik: I already did that on production.

db/schema.rb
tsubik: Dunno, new rails starts dumping this.

db/migrate/20240423144714_remove_observer_name_translation.rb
tsubik: There was only a small difference for one observer, and Sophie said to take English name for it.

PR #489, opened 2024-04-29
FMU name should not be translated

tsubik:
As in the title.

PT: https://www.pivotaltracker.com/story/show/187500588


santostiago (approved):
It looks good!

PR #490, opened 2024-04-30
Feature/operator documents new notifications admin

tsubik:
Send notification to operator's responsible admins about new pending document ready for review.

PT: https://www.pivotaltracker.com/story/show/187126374

### DIFF
--- a/app/resources/concerns/operator_documentable.rb
+++ b/app/resources/concerns/operator_documentable.rb
@@ -48,6 +48,11 @@ module OperatorDocumentable
 
     def attachment=(attachment)
       @model.build_document_file(attachment: attachment)
+      # TODO: check if there is better workaround for this issue https://github.com/rails/rails/issues/49898
+      # flipping activerecord flag about commit on first saved instance does not work
+      # when using build_document_file probably because of autosaving - save is called twice and in after commit we lose saved_changes hash somehow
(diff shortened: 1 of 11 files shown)


app/resources/concerns/operator_documentable.rb
```diff
@@ -48,6 +48,11 @@ def attachment
 
     def attachment=(attachment)
       @model.build_document_file(attachment: attachment)
+      # TODO: check if there is better workaround for this issue https://github.com/rails/rails/issues/49898
```
santostiago, line 51: 😮

config/locales/fr.yml
```diff
@@ -383,7 +383,7 @@ fr:
       annexes: 'Annexes'
       operator_document_id: "ID de document de l'opérateur"
       with_deleted: 'Avec supprimé'
-      reason_label: "Raison de mettre le document comme “non applicable”"
+      reason_label: "Le document a été mis comme non applicable avec l'explication suivante"
```
santostiago, line 386: 😆

santostiago (approved):
👍

app/resources/concerns/operator_documentable.rb
tsubik: Yeah, using too many AR callbacks could be a real trap in Rails project. Many times callbacks didn't work as expected.

PR #491, opened 2024-05-17
bundle update

tsubik:


.rubocop.yml
```diff
@@ -55,3 +55,7 @@ Rails/HasManyOrHasOneDependent:
 Rails/HasAndBelongsToMany:
   Enabled: false
 
+# have a mixed feeling about this style
+Rails/WhereRange:
```
santostiago, line 59: it was recommended somewhere?
If the query is simple and just the `>=` it's fine. If it has more conditions it might complicate things. But it can be ignored when it does.

santostiago (approved):

PR #496, opened 2024-06-19
Chore/refactor user permissions

tsubik:
Quick change for user permissions to not have to update_permissions for every user after some code change (that could be forgotten).

We only use managed permissions for backoffice managers, that's why it's better to use static `role_permissions` and not something saved in the database. In the database we will only save permissions overwrites, so all users except bo_managers will for now on have permissions as `nil` in the database.


santostiago (approved):
Cool!

PR #497, opened 2024-06-20
Add data normalization

tsubik:
Add data normalization to prevent issues like this (https://www.pivotaltracker.com/story/show/186906441) in the future.


santostiago (approved):
👍

PR #498, opened 2024-06-21
Unset application ENV variables when running deploy commands

tsubik:
Env variables were leaked into frontend apps when deploying using dashboard "Deploy" buttons.

More info in this issue here: https://www.pivotaltracker.com/story/show/187836380


santostiago (approved):
I read the description, but I'm not sure I got the issue.
Anyway, it won't hurt.

PR #501, opened 2024-08-13
Feature/two levels of observation qc new model

tsubik:
Changing observation QC flow to include two QC levels (QC1, QC2) instead of just one.

- QC reviewers for observers should be selectable, for both levels (observer responsible_for_qc1, responsible_for_qc2)
- QC should be available in observation tool as well (changing API to make it possible)
- admin panel - only QC2 for now 
- new Quality Control model for better view for whole QC process (Quality control model done as polymorphic to allow the same qc process for operator document - not sure if that's a good way to go)
- QC can be done only by selected users (not even admin role to prevent wrong persons doing QC - too many users with admin roles)
- Removing batch actions for start qc and ready for publication (QC should always be done individually, not in batch)
- Removing extra actions "Needs revision", "Ready for publication". All QC actions should be done through QC form
- Adding new user_permissions specs (we need a way more specs there, but it is outside of scope of this PR)


santostiago (approved):
I'm approving it to unblock it. I'll review it properly later.

PR #505, opened 2024-09-18
remove logo from observer

tsubik:
PT: https://www.pivotaltracker.com/story/show/188251197


santostiago (approved):
The pivotal task title says to remove the logos, but the description seems to be like "check why it's not being displayed".

tsubik:
Forgot to change description we talked with Sophie and during the meeting
decided that it's not needed and used anywhere, so better to remove it.

PR #510, opened 2024-10-17
Chore/refactor

tsubik:
Couple simple refactors
- remove interactor gem, and move code to either the controller for one interactor, or the separate PORO model in case of the other one,
- remove valid_url gem as looks like it does not work, instead simple custom URL validator
- remove some old single use rake tasks
- remove not needed initlializers, monkey patches as we already on rails 7.1
- remove validate_enum_attribute as Rails 7.1 has this built in with `validate` option  
- resolving some of TODO items


config/initializers/rgeo.rb
```diff
@@ -1,12 +0,0 @@
-# frozen_string_literal: true
-
-# Monkey patches this method (it's super slow)
-module RGeo
```
tsubik, line 4: @santostiago do you remember what was that for? Maybe it is not slow anymore, and anyway I don't think that a good reason for that kind of monkey patch, not sure what that method does, but that patching could have some unintentional consequences, that's why I'm removing it.

app/models/contributor.rb
```diff
@@ -27,4 +27,5 @@ class Contributor < ApplicationRecord
 
   validates :priority, numericality: {only_integer: true, greater_than_or_equal_to: 0}, if: :priority?
   validates :name, presence: true
+  validates :website, url: true
```
santostiago, line 30: we no longer have the gem, but shouldn't we validate the website?

app/models/operator_document_filter_tree.rb
```diff
@@ -0,0 +1,129 @@
...
+      {key: key, id: value[:index], name: value[:label]}
+    end.sort_by { |x| x[:name] }
+  end
+
+  def fmu_ids
+    Fmu.order(:name).pluck(:id, :name).map { |x| {id: x[0], name: x[1]} }
+  end
+
+  def required_operator_document_ids
+    # TODO: after fixing bad data remove that filter
```
santostiago, line 60: Is this still needed?

app/resources/v1/observation_resource.rb
```diff
@@ -13,7 +13,7 @@ class ObservationResource < BaseResource
       :litigation_status, :location_accuracy, :lat, :lng, :country_id,
       :fmu_id, :location_information, :subcategory_id, :severity_id,
       :created_at, :updated_at, :actions_taken, :validation_status, :validation_status_id,
-      :is_physical_place, :complete, :hidden, :user_type, :monitor_comment, :locale
+      :is_physical_place, :hidden, :user_type, :monitor_comment, :locale
```
santostiago, line 16: we no longer show `complete` on the FE?

santostiago (approved):
Good refactoring! 👍 🚀

app/admin/observer.rb
```diff
@@ -23,8 +23,7 @@ def scoped_collection
     column :is_active
     column :public_info
     column :countries do |observer|
-      names = observer.countries.map { |c| c.name }
-      names.join(" ").tr(",", ";")
+      observer.countries.map(&:name).join(";")
```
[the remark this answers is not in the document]
tsubik: Yes, it's not, but previous output was weird, countries names separated by space is weird and that change of comma to semicolon?

app/controllers/v1/operator_document_histories_controller.rb
```diff
@@ -3,16 +3,31 @@
 module V1
   class OperatorDocumentHistoriesController < APIController
     skip_before_action :authenticate, only: [:index, :show]
+    before_action :verify_filters, only: :index
     load_and_authorize_resource class: "OperatorDocumentHistory"
 
-    def index
-      result = SearchDocumentInTime.call(params)
+    FilterError = Class.new(StandardError)
```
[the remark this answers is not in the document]
tsubik: Eventually if needed in more classes then Error could be moved up to APIController.

app/models/operator_document_filter_tree.rb
tsubik: I checked and there are still some required operator documents without country. I will have to ask Sophie about it and come back to this in some future refactor.

app/resources/v1/observation_resource.rb
tsubik: No, I checked portal and obs tool code, didn't find that attribute.

PR #513, opened 2024-11-20
Feature/Observation with non concession activity

tsubik:
Ability to note that producer is doing activities in FMU that it has no official concession (non_concession_activity). In this case, it would be possible to select non operator FMUs in the forms.

This feature is only available for DR Congo observations. 

We will not show FMU name in the portal for those observations. 

PT: https://www.pivotaltracker.com/story/show/187336758


santostiago (approved):
It looks fine.
They won't need documents, right?

tsubik:
Not sure what you mean @santostiago, which documents? This is only to be able to select FMU not belonging to the producer.

PR #525, opened 2025-02-28
Improve FMUs show page

tsubik:
PT: https://www.pivotaltracker.com/story/show/188819294

- two column layout
- view geojson in modal dialog
- view geojson properties in modal dialog
- [fix] show shapefiles sidebar only on the index page

new_fmu_show_page


santostiago (commented):
The code is good. Let me just test it.

santostiago (approved):
It looks fine.
But I see the `x` closing button on the left side and it doesn't work.
Am I missing some JS library?

santostiago:
Screenshot 2025-03-30 at 19 30 29

tsubik:
Hmm, that looks weird on your end, like stylesheets are not refreshed somehow. Did you try to run `rails assets:clobber` or `tmp:clear`?

tsubik:
I pushed this to staging and also looks good to me.

PR #592, opened 2026-04-16
Chore/paper trail migration to jsonb columns

tsubik:
Paper trail migration to JSONB columns for better performance and less problems in the future.


db/schema.rb
```diff
@@ -1132,10 +1132,12 @@
     t.bigint "item_id", null: false
     t.string "event", null: false
     t.string "whodunnit"
-    t.text "object"
+    t.text "old_object"
```
santostiago, line 1136: This was a text field? 🤯

lib/tasks/data_migrations.rake
```diff
@@ -50,4 +50,37 @@ namespace :data_migrations do
     end
     puts "Wrote tmp/data_migrations_report_mission_type_report.csv"
   end
+
+  desc "Move PaperTrail versions from YAML to JSON columns."
```
santostiago, line 54: Oh.. it was YAML. Then text field made sense.

lib/tasks/data_migrations.rake
```diff
@@ -50,4 +50,37 @@ namespace :data_migrations do
...
+    errors = 0
+    puts "Migrating #{total} PaperTrail versions to JSON (batch_size=#{batch_size})..."
+
+    scope.find_in_batches(batch_size: batch_size) do |batch|
+      ActiveRecord::Base.transaction do
+        batch.each do |version|
+          attrs = {}
+          attrs[:object] = PaperTrail::Serializers::YAML.load(version.old_object) if version.old_object.present?
+          attrs[:object_changes] = PaperTrail::Serializers::YAML.load(version.old_object_changes) if version.old_object_changes.present?
+          version.update_columns(attrs) if for_real
```
santostiago, line 72: Mmmm.. ok... but this just keeps the old data in a new field.
Shouldn't we also migrate the old data into JSON and start using that field everywhere?

santostiago (approved):

lib/tasks/data_migrations.rake
tsubik: YAML loading will load to hash, and that will be saved in new object and object_changes fields as JSON.

tsubik: Then in the next PR, I will remove old fields.

PR #593, opened 2026-04-16
Observation: add reject action for published observations

tsubik:
Allow admin users to easily reject published observations.

image-20260413-112210

JIRA: https://gfw.atlassian.net/browse/OPEN-379

### DIFF
--- a/app/admin/observation.rb
+++ b/app/admin/observation.rb
@@ -70,7 +70,7 @@ ActiveAdmin.register Observation do
   end
 
   member_action :reject, method: [:get, :post] do
-    unless can?(:reject, resource)
+    unless resource.published? && current_user.admin?
       flash[:notice] = I18n.t("active_admin.observations_page.observation_cannot_be_rejected")
       render js: "window.location.reload();" and return
     end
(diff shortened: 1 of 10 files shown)


spec/controllers/admin/observations_controller_spec.rb
```diff
@@ -42,6 +42,61 @@
...
+      describe "GET reject" do
+        let(:observation) { create(:observation) }
+
+        before { get :reject, params: {id: observation.id}, xhr: true, format: :js }
+
+        it "renders the reject form js" do
+          expect(response).to be_successful
+        end
+
+        context "when observation is not published and user is not admin" do
```
santostiago, line 55: if you want to test both branches of the condition, you should add a test for when the user is admin and the observation is published.

app/admin/observation.rb
```diff
@@ -69,6 +69,32 @@ def scoped_collection
     end
   end
 
+  member_action :reject, method: [:get, :post] do
+    unless resource.published? || current_user.admin?
```
santostiago: wouldn't it be better to use an ability for this?

santostiago (approved):

app/admin/observation.rb
tsubik: Ok, good idea.

spec/controllers/admin/observations_controller_spec.rb
tsubik: That's the default test couple lines above. I will add a better test description.

app/admin/observation.rb
tsubik: Actually, I'm reverting to previous behaviour as changing abilities is too complicated now. We need to address that in separate PR. There are things like manage: :all for admins that is making my new ability for reject only published not working.

PR #595, opened 2026-04-24
Hot fix/observation rejected qc2

tsubik:
Allow observation state transition from Rejected to Ready for QC2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -64,7 +64,7 @@ jobs:
 
       - name: Install other dependencies
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
       - name: Install other dependencies
         run: |
-          npm install -g mjml
+          npm install -g mjml@4
```
santostiago, line 67: yeah, this caused me troubles

santostiago (approved):

PR #596, opened 2026-04-30
Notify users that they haven't login in a while and deactivate accounts

santostiago:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: https://gfw.atlassian.net/browse/OPEN-376

## What is the new behavior?
We run a job every day that warns users that haven't logged in in a while that their accounts will be disabled.
Disable users that have not logged in in 2 years.

- [ ] Yes
- [x] No

### DIFF
--- a/config/locales/mailers.en.yml
+++ b/config/locales/mailers.en.yml
@@ -24,6 +24,14 @@ en:
       message: Please click on the link to reset your password to connect to the Open Timber Portal
       link_expiration: The above link will expire in 6 hours. If you made multiple requests to reset your password, please make sure to click on the link in the most recent e-mail.
       outro: If you didn't request this password reset, please ignore this email. Your password will not change until you access the link above and create a new password.
+    inactive_account_warning:
+      subject: "Open Timber Portal account inactivity notice - disable date %{disable_date}"
+      message: "Your account has not been used for 18 months. Please log in soon to keep it active. If there is no login activity, your account will be disabled on %{disable_date}."
+      cta: "Log in here: %{link}"
+    account_deactivated_for_inactivity:
+      subject: "Your Open Timber Portal account has been deactivated"
+      message: "Because your account has not been used for 2 years, it is now deactivated."
+      contact: "If you need help or think this is an error, please contact the OTP team."
   observation_mailer:
     observation_created:
       subject: Observation %{id} was successfully created
--- a/lib/tasks/scheduler.rake
+++ b/lib/tasks/scheduler.rake
@@ -99,4 +99,44 @@ namespace :scheduler do
     Rails.logger.info "Rails cache cleaned. It took #{time} ms."
     Rails.logger.info "::::::::::::::::::::::::::::::::::::::::::::::::::::::::::"
   end
+
+  desc "Warn and deactivate inactive users"
+  task deactivate_inactive_users: :environment do
+    Rails.logger.info "::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::"
+    Rails.logger.info "Going to process inactive users at: #{Time.zone.now.strftime("%d/%m/%Y %H:%M")}"
+
+    warning_start = 18.months.ago.beginning_of_day
+    warning_end = 18.months.ago.end_of_day
+    deactivation_threshold = 2.years.ago.end_of_day
+
+    warned_users = 0
+    deactivated_users = 0
+
+    time = Benchmark.ms do
+      warning_users = User.where(is_active: true)
+        .where("COALESCE(last_sign_in_at, created_at) BETWEEN ? AND ?", warning_start, warning_end)
+
+      warning_users.find_each do |user|
+        disable_date = (user.last_sign_in_at || user.created_at).to_date + 2.years
+        I18n.with_locale(user.locale.presence || I18n.default_locale) do
+          UserMailer.inactive_account_warning(user, disable_date).deliver_now
+        end
+        warned_users += 1
+      end
+
+      users_to_deactivate = User.where(is_active: true)
+        .where("COALESCE(last_sign_in_at, created_at) <= ?", deactivation_threshold)
+
+      users_to_deactivate.find_each do |user|
+        user.update!(is_active: false, deactivated_at: Time.zone.now)
+        I18n.with_locale(user.locale.presence || I18n.default_locale) do
+          UserMailer.account_deactivated_for_inactivity(user).deliver_now
+        end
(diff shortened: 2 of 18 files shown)


spec/lib/tasks/deactivate_inactive_users_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  after(:each) do
+    Rake::Task["scheduler:deactivate_inactive_users"].reenable
+    ActionMailer::Base.deliveries.clear
+  end
+
+  subject(:run_task) { Rake::Task["scheduler:deactivate_inactive_users"].invoke }
+
+  around do |example|
+    travel_to(Time.zone.parse("2026-04-30 10:00:00")) { example.run }
```
tsubik, line 16: Is it necessary to set fixed time here for every example?

config/locales/mailers.en.yml
```diff
@@ -22,6 +22,14 @@ en:
...
       link_expiration: The above link will expire in 6 hours. If you made multiple requests to reset your password, please make sure to click on the link in the most recent e-mail.
       outro: If you didn't request this password reset, please ignore this email. Your password will not change until you access the link above and create a new password.
+    inactive_account_warning:
+      subject: "Open Timber Portal account inactivity notice - disable date %{disable_date}"
+      message: "Your account has not been used for 18 months. Please log in soon to keep it active. If there is no login activity, your account will be disabled on %{disable_date}."
+      cta: "Log in here: %{link}"
+    account_deactivated_for_inactivity:
+      subject: "Your Open Timber Portal account has been deactivated"
+      message: "Because your account has not been used for 2 years, it is now deactivated."
+      contact: "If you need help or think this is an error, please contact the OTP team."
```
tsubik, line 34: Should we put `CONTACT_EMAIL` here as well? I think question to Sophie.

lib/tasks/scheduler.rake
```diff
@@ -90,4 +90,47 @@ namespace :scheduler do
...
+
+    warning_threshold = 18.months.ago.end_of_day
+    warning_cooldown = 1.month.ago.end_of_day
+    deactivation_threshold = 2.years.ago.end_of_day
+
+    warned_users = 0
+    deactivated_users = 0
+
+    time = Benchmark.ms do
+      warning_users = User.where(is_active: true)
```
tsubik: We should filter out special user `webuser@example.com`, that user cannot be deactivated.

lib/tasks/scheduler.rake
```diff
@@ -90,4 +90,47 @@ namespace :scheduler do
...
+      warning_users.find_each do |user|
+        disable_date = (user.last_sign_in_at || user.created_at).to_date + 2.years
+        I18n.with_locale(user.locale.presence || I18n.default_locale) do
+          UserMailer.inactive_account_warning(user, disable_date).deliver_now
+        end
+        user.update!(last_inactivity_warning_sent_at: Time.zone.now)
+        warned_users += 1
+      end
+
+      users_to_deactivate = User.where(is_active: true)
```
tsubik: I wonder should we at least warn user once? Or just go straight to deactivate. The problem is only for the first run. I see a lot of users will be deactivated.

tsubik (changes requested):
One thing should be changed, webuser cannot be deactivated. Other than that, looks really good.

lib/tasks/scheduler.rake
santostiago: Sophie said it should be fine.

config/locales/mailers.en.yml
santostiago: I think it's not needed. They can just login again.

spec/lib/tasks/deactivate_inactive_users_spec.rb
santostiago: you mean to use just the date?

tsubik: No, I just don't understand why to set time for each example as in tests relative time is used.

lib/tasks/scheduler.rake
```diff
@@ -90,4 +90,52 @@ namespace :scheduler do
...
+    deactivated_users = 0
+
+    disablable_users = User.where(is_active: true)
+      .where.not(email: "webuser@example.com")
+      .joins(:user_permission)
+      .where.not(user_permissions: {user_role: %w[ngo ngo_manager] })
+
+    time = Benchmark.ms do
+      warning_users = disablable_users
+        .where("COALESCE(last_sign_in_at, created_at) <= ?", warning_threshold)
```
tsubik: After adding join to `user_permission` there is an error of ambiguous column, should be `users.created_at`. The same in 2 other places below.

tsubik (commented):
Small issue with the script.

lib/tasks/scheduler.rake
santostiago: Good catch! (as I said, I didn't test it... I'll clone the DB so that I can run it - or maybe there should be tests for this file).

spec/lib/tasks/deactivate_inactive_users_spec.rb
santostiago: As a principle I try to freeze the time when running time-dependent tests. But, yes, it would work without it.

tsubik (changes requested):
I found some edge case that we should consider before merging this. What if the user is deactivated, informed, ask about activation and admin will activate that user, but the user will not take any action in the next 24hrs, or less depends when activation happens. That user will be deactivated once again. I think we need to give users couple days to log in before deactivating again (save reactivated_at date? or maybe other solution)

santostiago:
> I found some edge case that we should consider before merging this. What if the user is deactivated, informed, ask about activation and admin will activate that user, but the user will not take any action in the next 24hrs, or less depends when activation happens. That user will be deactivated once again. I think we need to give users couple days to log in before deactivating again (save reactivated_at date? or maybe other solution)

I'm not sure if we should give them a grace period to log in again, or if we should just mark the `last_login_at` to be now when enabling a user in the Backoffice.
I don't think this is very important, but I'll add a new field so that we have the whole history.

santostiago:
I'm going to rebase it.

tsubik (approved):
Looks good🚀

tsubik:
I configured a new check in healthchecks and now it's ready to deploy.

PR #605, opened 2026-05-21
observations: move obsevations to qc2 when all qc1 reviewers removed

tsubik:
Moving all QC1 observations to QC2 if there is no QC1 reviewer left.

JIRA: https://gfw.atlassian.net/browse/OPEN-388

### DIFF
--- a/app/models/observer.rb
+++ b/app/models/observer.rb
@@ -56,6 +56,7 @@ class Observer < ApplicationRecord
   validates :data_email, format: {with: EMAIL_VALIDATOR, if: :data_email?}
 
   before_create :set_responsible_qc2
+  after_update :move_qc1_observations_to_qc2, if: -> { saved_change_to_responsible_qc1_id?(to: nil) }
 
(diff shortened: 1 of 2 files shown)


app/models/observer.rb
```diff
@@ -56,6 +56,7 @@ class Observer < ApplicationRecord
   validates :data_email, format: {with: EMAIL_VALIDATOR, if: :data_email?}
 
   before_create :set_responsible_qc2
+  after_update :move_qc1_observations_to_qc2, if: -> { saved_change_to_responsible_qc1_id?(to: nil) }
```
santostiago, line 59: And what happens if there is no QC1 reviewer to start with?

santostiago (approved):

app/models/observer.rb
tsubik: Then nothing happens. There should not be any QC1 observations as those are going to this state only if there is that person.

PR #606, opened 2026-05-21
Fix/long running cache invalidation

tsubik:
Sometimes that cache invalidation takes too long, moving to background job. Run cache cleanup every day and set expiration time to 7 days.


santostiago (approved):
Was this a problem?

tsubik:
Yes, there was a very weird issue, because that created some racing condition when approving the publication authorization document.

PR #609, opened 2026-05-27
Remove api keys, refactor API to not need a webuser

tsubik:
So far, API keys are just adding unnecessary friction to the API and not actually doing anything useful. If we need something of this sort in the future, we can rebuild it quickly in a better way.


santostiago (approved):
Nice work.
This was added at the very beginning but never took off.

PR #610, opened 2026-06-03
Feature/api auth refactor

tsubik:
Add cookie-based auth to API. Frontend apps are using the same domain, and Angular app is SPA and cookies are the most secure way of storing user authentication.

JIRA: https://gfw.atlassian.net/browse/OPEN-389

### DIFF
--- a/app/resources/concerns/operator_documentable.rb
+++ b/app/resources/concerns/operator_documentable.rb
@@ -114,7 +114,10 @@ module OperatorDocumentable
 
   module ClassMethods
     def apply_includes(records, directives)
-      super.includes(:document_file, :required_operator_document)
+      result = super
+      return result unless result.respond_to?(:includes)
+
+      result.includes(:document_file, :required_operator_document)
     end
   end
 end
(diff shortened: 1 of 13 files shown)


app/resources/concerns/operator_documentable.rb
```diff
@@ -83,7 +83,10 @@ def can_see_document?
 
   module ClassMethods
     def apply_includes(records, directives)
-      super.includes(:document_file, :required_operator_document)
+      result = super
```
santostiago, line 117: This has nothing to do with the PR's goal, right?

santostiago (approved):
It all seems fine, but what is the goal?
What is the problem with the existing authorization header?

app/resources/concerns/operator_documentable.rb
tsubik: Yes, one of those methods was failing somewhere; that's why I added that safeguard everywhere.

tsubik:
@santostiago the goal is to eliminate high security issues and simplify things. I had an alternative: install doorkeeper or reimplement OAuth, as the current token system was not great (tokens not expiring, cannot be revoked, there should be short- and long-lived tokens, refresh tokens, the Angular app is an SPA - no server and it was keeping tokens in LocalStorage - bad practice). I decided that cookies as a standard for web-based applications are the safest and easiest choice. I'm currently using cookies for authorized download sessions, and with this I can remove download sessions and just keep this.

PR #625, opened 2026-07-15
Provision script, terraform and documentation updates

tsubik:
- fix autopostgresbackup config
 - enable intelligent tiering for s3 buckets
 - add option to disable cron sync (like for staging we don't actually need that after it is tested - objects can be manually removed from the bucket as well)
 - updating documentation to better reflect architecture and usage of scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,13 @@
 # OPEN TIMBER PORTAL API #
 
+Ruby on Rails backend for the [Open Timber Portal](https://opentimberportal.org),
+serving the public portal and the observations tool. It exposes a JSON API, an
+admin interface, and background processing (Sidekiq) for imports, notifications,
+and scheduled jobs.
+
+- **Local development:** [Docker](#using-docker) or [natively](#natively)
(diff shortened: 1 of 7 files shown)


santostiago (approved):
Yeah, I don't know how terraform works :D 
But the backups are being created on the same EC2 machine? Shouldn't they be in S3?
Also, does it make sense to have a managed DB instead of having postgres running on the same machine as the apps? I know it's been like this since ever, but it can make the backups smaller (that's pretty much the only advantage, because we won't need to scale).

tsubik:
We can add managed DB and managed Redis; there are some benefits, but also higher costs for extra instances. Backups are synced to S3.

tsubik:
For now, I just recreated the current infrastructure in code. We can iterate on this. I would say, more benefits would come from adding Kamal for Docker deployments and installing redis, postgres still on the same machine in Docker environments. But that needs time to figure out the configuration with NextJs and Angular deployments, along with the main app.

PR #627, opened 2026-07-22
Feature/New DRC concession types

tsubik:
This PR is adding a new concession type for DRC. There were no explicit types before, so existing FMUs will gain `ccf` type, and admin can change the type later through backoffice. That is why we have to enable changing the forest type for FMU and that will change the required documents as well. 

JIRA: https://gfw.atlassian.net/browse/OPEN-395

### DIFF
--- a/app/models/fmu_operator.rb
+++ b/app/models/fmu_operator.rb
@@ -114,7 +114,7 @@ class FmuOperator < ApplicationRecord
 
       # Only the RODF for this fmu's forest_type should be created, empty forest_types means the document applies to all FMUs
       required_documents = RequiredOperatorDocumentFmu.where(
-        "country_id = :country_id AND (forest_types = '{}' OR :forest_type = ANY (forest_types))",
+        "country_id = :country_id AND (COALESCE(cardinality(forest_types), 0) = 0 OR :forest_type = ANY (forest_types))",
         country_id: fmu.country_id, forest_type: Fmu.forest_types[fmu.forest_type]
       )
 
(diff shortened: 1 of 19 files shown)


app/models/fmu_operator.rb
```diff
@@ -93,39 +93,54 @@ def self.calculate_current
...
-  # Updates the list of documents for this FMU, TODO: move it to some service object
+  # Updates the list of documents for this FMU
   def update_documents_list
-    current_operator = self&.fmu&.reload&.operator
+    return if fmu_id.blank?
+
+    current_operator = fmu.reload.operator
 
     OperatorDocumentFmu.transaction do
+      # FMUs in non-active countries use the generic required documents (country_id is null)
```
santostiago, line 103: we're not expecting that old FMUs change their type, right?
Probably only new ones will be of the new type.
Otherwise we're risking deleting old docs which we might not want.

lib/tasks/data_migrations.rake
```diff
@@ -83,4 +83,29 @@ namespace :data_migrations do
     warn "\n  #{errors} version(s) failed to parse and were skipped." if errors > 0
     puts "\nFinished migrating PaperTrail versions to JSON columns. You can now remove the old YAML columns with a separate migration."
   end
+
+  desc "Set forest_types to CCF on existing COD required operator documents and migrate COD FMUs to the CCF forest type."
```
santostiago, line 87: This forest type exists only for COD?

santostiago (approved):

app/models/fmu_operator.rb
tsubik: That is exactly a requirement: to change the FMU type and archive existing documents.

lib/tasks/data_migrations.rake
tsubik: Yes

tsubik: Right now, all have default FMU type, but there will be 2 types for COD, and the default for COD is CCF.

PR #630, opened 2026-07-26
[OPEN-382] Enable Devise account lockout after failed logins

santostiago:
## Summary
- Enable Devise `:lockable` so accounts lock after 10 failed sign-in attempts
- Unlock via email link or automatically after 1 hour
- Wire lockout into the API `/login` endpoint (it previously bypassed Devise) and cover lock, unlock email, and unlock-link flows in specs

## Test plan
- [ ] Run `bundle exec rspec spec/integration/v1/sessions_spec.rb`
- [ ] Confirm failed API logins increment `failed_attempts` and lock after 10 attempts
- [ ] Confirm unlock instructions email is sent on lock and `/admin/unlock?unlock_token=…` unlocks the account
- [ ] Confirm successful login resets `failed_attempts`
- [ ] Confirm ActiveAdmin login is also blocked for locked users

Made with [Cursor](https://cursor.com)


santostiago:
When my other PR gets merged I'll rebase this one and the brakeman issue should be solved.

tsubik:
There is a small issue with extra padding on the backoffice login page

image

tsubik (changes requested):
Looks good for the back office, only small issues. We also need to customize this for frontend apps. Not sure if in the same way as forgotten password, that is a lot of extra code to customize emails, or maybe there is a better way to keep the email but customize redirect after unlock - to not end up on backoffice log in page when operator account clicked unlock link.

tsubik:
Or maybe we can start with only the `time` strategy and allow admins to unlock users? Would be good to add some locked? information with a date to the user's show page as well.

tsubik (approved):
👍

